PR #4, opened 2019-04-09
Add NdNorms chunk

danieldk:
This chunk stores the norms of in-vocabulary words. If an Embedding
data structure stores NdNorms, the vector norms can be queried with
the `embedding_with_norm` method.

When converting from word2vec/text in `ff-convert`, always normalize
embeddings and store norms.

Some random points:

* I still need to add some more unit tests.
* The norms chunk is currently not a generic type of `Embedding`. This is to keep the type signature simple while we still can. This may change later. It will definitely be necessary when we want to make it memory mappable.
* The norms chunk is optional, for compatibility with existing finalfusion embeddings. Since we can assume the (known) *word* embeddings to be normalized, the norm 1 is returned when the chunk is absent.
* `ff-convert` from word2vec and text now always normalizes the embeddings and creates a norms chunk.
* I am considering to multiply the unit vectors by their norms when converting *to* word2vec or text embeddings, to restore the raw, unnormalized embeddings. However, this is currently not done.
* I tried to make the norms mandatory (normalizing the embeddings and computing norms when the chunk is not present). However, this poses problems, because mmaped and quantized embeddings cannot be updated.
* The traits for reading word2vec/text embeddings still have an argument for normalization. These are still needed for roundtrip unit tests. I might make a crate-private trait for unnormalized reading and call this from the public trait that always normalizes. This would allow us to keep the roundtrip unit tests, while always normalizing in the public interface. (Done now for `ReadWord2Vec`)

### DIFF
--- a/finalfusion/src/embeddings.rs
+++ b/finalfusion/src/embeddings.rs
@@ -8,14 +8,19 @@ use std::slice;
 
 use failure::{ensure, Error};
 use ndarray::Array1;
+use rand::{FromEntropy, Rng};
+use rand_xorshift::XorShiftRng;
+use reductive::pq::TrainPQ;
 
 use crate::io::{
     private::{ChunkIdentifier, Header, MmapChunk, ReadChunk, WriteChunk},
     MmapEmbeddings, ReadEmbeddings, WriteEmbeddings,
 };
 use crate::metadata::Metadata;
+use crate::norms::{NdNorms, Norms};
 use crate::storage::{
-    CowArray, CowArray1, MmapArray, NdArray, QuantizedArray, Storage, StorageViewWrap, StorageWrap,
+    CowArray, CowArray1, MmapArray, NdArray, Quantize as QuantizeStorage, QuantizedArray, Storage,
+    StorageView, StorageViewWrap, StorageWrap,
 };
 use crate::util::l2_normalize;
 use crate::vocab::{SimpleVocab, SubwordVocab, Vocab, VocabWrap, WordIndex};
@@ -30,21 +35,48 @@ pub struct Embeddings<V, S> {
     metadata: Option<Metadata>,
     storage: S,
     vocab: V,
+    norms: Option<NdNorms>,
 }
 
-impl<V, S> Embeddings<V, S> {
-    /// Construct an embeddings from a vocabulary and storage.
-    pub fn new(metadata: Option<Metadata>, vocab: V, storage: S) -> Self {
+impl<V, S> Embeddings<V, S>
+where
+    V: Vocab,
+{
+    /// Construct an embeddings from a vocabulary, storage, and norms.
+    ///
+    /// The embeddings for known words **must** be
+    /// normalized. However, this is not verified due to the high
+    /// computational cost.
+    pub fn new(metadata: Option<Metadata>, vocab: V, storage: S, norms: NdNorms) -> Self {
+        assert_eq!(
+            vocab.len(),
+            norms.0.len(),
+            "Vocab and norms do not have the same length"
+        );
+
         Embeddings {
             metadata,
             vocab,
             storage,
+            norms: Some(norms),
+        }
+    }
+}
+
+impl<V, S> Embeddings<V, S> {
+    pub(crate) fn new_without_norms(metadata: Option<Metadata>, vocab: V, storage: S) -> Self {
+        Embeddings {
+            metadata,
+            vocab,
+            storage,
+            norms: None,
         }
     }
 
-    /// Decompose embeddings in its vocabulary and storage.
-    pub fn into_parts(self) -> (Option<Metadata>, V, S) {
-        (self.metadata, self.vocab, self.storage)
+    /// Decompose embeddings in its vocabulary, storage, and
+    /// optionally norms.
+    pub fn into_parts(self) -> (Option<Metadata>, V, S, Option<NdNorms>) {
+        (self.metadata, self.vocab, self.storage, self.norms)
     }
 
     /// Get metadata.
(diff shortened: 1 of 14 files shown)


sebpuetz:
I'm not done going through all the changes yet.

> * The norms chunk is optional, for compatibility with existing finalfusion embeddings. Since we can assume the (known) word embeddings to be normalized, the norm 1 is returned when the chunk is absent.
> * ff-convert from word2vec and text now always normalizes the embeddings and creates a norms chunk.

With these two points, the only way to get unnormalized finalfusion embeddings would be through Rust. Perhaps, we can only expose a constructor in the public API with norms instead of doing that optionally? Then we'd be fairly sure that no finalfusion file without a norms chunk but unnormalized embeddings floats around.

> * I am considering to multiply the unit vectors by their norms when converting to word2vec or text embeddings, to restore the raw, unnormalized embeddings. However, this is currently not done.

Could be optional? Although that could be messy without subcommands which, iirc, you didn't like.

> * The traits for reading word2vec/text embeddings still have an argument for normalization. These are still needed for roundtrip unit tests. I might make a crate-private trait for unnormalized reading and call this from the public trait that always normalizes. This would allow us to keep the roundtrip unit tests, while always normalizing in the public interface. (Done now for ReadWord2Vec)

Sounds reasonable to me.

finalfusion/src/embeddings.rs
```diff
@@ -104,6 +131,37 @@ where
         }
     }
 
+    /// Get the embedding and original norm of a word.
+    ///
+    /// Returns for a:
```
sebpuetz: Somethings missing here?

finalfusion/src/norms.rs
```diff
@@ -0,0 +1,140 @@
+use std::io::{Read, Seek, SeekFrom, Write};
+use std::mem::size_of;
+
+use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
+use failure::{ensure, format_err, Error};
+use ndarray::Array1;
+
+use crate::io::private::{ChunkIdentifier, ReadChunk, TypeId, WriteChunk};
+
+pub trait Norms {
```
sebpuetz: Final version of the PR would also need docs for the norms module

finalfusion/src/storage.rs
```diff
@@ -687,6 +687,18 @@ impl StorageView for StorageViewWrap {
     }
 }
 
+/// Storage that provide a view of the embedding matrix.
```
sebpuetz: mutable view?

danieldk:
> With these two points, the only way to get unnormalized finalfusion embeddings would be through Rust. Perhaps, we can only expose a constructor in the public API with norms instead of doing that optionally? Then we'd be fairly sure that no finalfusion file without a norms chunk but unnormalized embeddings floats around.

Good idea. Done!

danieldk:
> Could be optional? Although that could be messy without subcommands which, iirc, you didn't like.

Why? It's just one option/flag.

danieldk:
Ok, this is ready for review. Since it is already a big change, I will make a separate PR for 'unnormalization' and perhaps another one for memory mapping norms.

sebpuetz:
> > Could be optional? Although that could be messy without subcommands which, iirc, you didn't like.
> 
> Why? It's just one option/flag. And we already had one for normalizing word2vec embeddings, which is now gone ;).

I think I was referring to having options only apply to a single conversion? But I guess it's possible to add conflicts with other conversions or maybe we can whitelist compatible options.

danieldk:
> I think I was referring to having options only apply to a single conversion? But I guess it's possible to add conflicts with other conversions or maybe we can whitelist compatible options.

Indeed, in those cases the option could be ignored or the program could fail with an error.

sebpuetz:
Rust stable CI job had failed, after manual restart it went through. I'm reviewing the changes, should be done soon-ish!

finalfusion/src/embeddings.rs
```diff
@@ -104,6 +142,37 @@ where
...
+    ///
+    /// * The word embedding.
+    /// * The norm of the embedding before normalization to a unit vector.
+    ///
+    /// The original embedding can be reconstructed by multiplying all
+    /// embedding components by the original norm.
+    ///
+    /// If the model does not have associated norms, *1* will be
+    /// returned as the norm for vocabulary words.
+    pub fn embedding_with_norm(&self, word: &str) -> Option<(CowArray1<f32>, f32)> {
```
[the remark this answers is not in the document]
danieldk: I think tuple structs only have very limited applicability (things like RGB values, coordinates), but I agree that this is somewhat difficult to read. So I made a POD `EmbeddingWithNorm`.

finalfusion/src/embeddings.rs
```diff
@@ -261,6 +444,27 @@ impl<'a> Iterator for Iter<'a> {
...
 
+/// Iterator over embeddings.
+pub struct IterWithNorms<'a> {
+    storage: &'a Storage,
+    norms: Option<&'a NdNorms>,
+    inner: Enumerate<slice::Iter<'a, String>>,
+}
+
+impl<'a> Iterator for IterWithNorms<'a> {
+    type Item = (&'a str, CowArray1<'a, f32>, f32);
```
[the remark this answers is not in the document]
danieldk: Thanks, uses `EmbeddingWithNorm` now as well!

danieldk:
On 12 Apr 2019, at 14:55, Sebastian Pütz <notifications@github.com> wrote:
> @sebpuetz approved this pull request.
> 
> Looks like I needed to re-approve after the force-push?

The admin hammer also works ;).

PR #9, opened 2019-06-08
Add parameters to mask analogy queries.

sebpuetz:
Addresses finalfusion/finalfusion-utils#9

The mask allows to include or exclude a, b or c in analogy queries like: a is to b like c is to d.

Seems like I don't have write access to the finalfusion repo, so this is coming from a fork.

### DIFF
--- a/src/similarity.rs
+++ b/src/similarity.rs
@@ -53,6 +53,30 @@ pub trait Analogy {
         word2: &str,
         word3: &str,
         limit: usize,
+    ) -> Option<Vec<WordSimilarity>> {
+        self.analogy_masked(word1, word2, word3, limit, [true, true, true])
+    }
+
+    /// Perform an analogy query.
+    ///
+    /// This method returns words that are close in vector space for the
+    /// analogy query `word1` is to `word2` as `word3` is to `?`. More
+    /// concretely, it searches embeddings that are similar to:
+    ///
+    /// *embedding(word2) - embedding(word1) + embedding(word3)*
+    ///
+    /// At most, `limit` results are returned.
+    ///
+    /// `remove` specifies which parts of the queries are excluded from the
+    /// output candidates. If `remove[0]` is `true`, `word1` cannot be
+    /// returned as an answer to the query.
+    fn analogy_masked(
+        &self,
+        word1: &str,
+        word2: &str,
+        word3: &str,
+        limit: usize,
+        remove: [bool; 3],
     ) -> Option<Vec<WordSimilarity>>;
 }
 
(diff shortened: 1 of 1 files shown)


src/similarity.rs
```diff
@@ -54,6 +54,15 @@ pub trait Analogy {
         word3: &str,
         limit: usize,
     ) -> Option<Vec<WordSimilarity>>;
+
+    fn analogy_masked(
```
danieldk: The `analogy` method above could now provide a default implementation that calls `analogy_by` the `analogy` implementation in the `impl` block can then be removed.

src/similarity.rs
```diff
@@ -68,12 +77,31 @@ where
...
     ) -> Option<Vec<WordSimilarity>> {
-        self.analogy_by(word1, word2, word3, limit, |embeds, embed| {
-            embeds.dot(&embed)
-        })
+        self.analogy_by(
+            word1,
+            word2,
+            word3,
+            limit,
+            [false, false, false],
```
danieldk: I am a bit confused. I thought being masked would mean that the word is not included in the result set. Then the 'traditional' analogy implementation should have `[true, true, true]` as the mask?

src/similarity.rs
```diff
@@ -120,7 +150,12 @@ where
         let mut embedding = (&embedding2.as_view() - &embedding1.as_view()) + embedding3.as_view();
         l2_normalize(embedding.view_mut());
 
-        let skip = [word1, word2, word3].iter().cloned().collect();
+        let skip = [word1, word2, word3]
+            .iter()
+            .zip(mask.iter())
+            .filter(|(_, &include)| !include)
```
danieldk: See above, I would expect masking to exclude the words (*masking* in the *conceal* sense). Perhaps mask is not the correct verb to use, maybe *exclude*?

src/similarity.rs
sebpuetz: I initially had a failing CI build because I confused myself with this, too. 

I was thinking of masks as something like we use in tf models. true (or 1) for active steps and false (or 0) for inactive steps. Similarly here, false for those words that are inactive.

sebpuetz:
Pushed the changes

danieldk:
> Pushed the changes

Did you see the suggestion about the name change (masked -> remove)?

Also, there is no `analogy_by` with default masking.

sebpuetz:
> Did you see the suggestion about the name change (masked -> remove)?

Missed that.

> Also, there is no `analogy_by` with default masking.

I thought you meant a default implementation for `Analogy::analogy`, which I added. 

Now I also added another method `AnalogyBy::analogy_by_masked` and a default impl for `AnalogyBy::analogy_by` with the default masking. Also, rustdoc was missing for the new method, which I added now.

PR #14, opened 2019-07-09
Report which embeddings could be computed in failed analogy queries

danieldk:


danieldk:
> Looks good to me.
> 
> I'm wondering whether we want to return something per default indicating which representations were constructed from subword representations.
> 
> Currently a user has to implement that themselves by checking each component of the analogy against the vocabulary.

Although it is very implicit, you could actually get this by checking whether the `CowArray` is owned (subwords) or borrowed (known word).

sebpuetz:
> > Looks good to me.
> > I'm wondering whether we want to return something per default indicating which representations were constructed from subword representations.
> > Currently a user has to implement that themselves by checking each component of the analogy against the vocabulary.
> 
> Although it is very implicit, you could actually get this by checking whether the `CowArray` is owned (subwords) or borrowed (known word).

Tho that wouldn't give info on what parts were missing, just that something was missing.

danieldk:
> Tho that wouldn't give info on what parts were missing, just that something was missing.

Ah, you mean specifically in analogy queries? Indeed, that's currently not possible at all. Though I am not sure how often you'd want to know for an analogy query which word representations were constructed with subword units. It also complicates the API, because then you'd have to return 3 times an enum plus the actual analogy results.

sebpuetz:
Yeah, I think it's justifiable to put the burden on the caller, if anyone ever even wants to check that.

PR #15, opened 2019-07-11
Implement a library-specific error type

danieldk:
Design choices:

- `Error` is in the `io` module. We might want to introduce other types of errors when necessary, but we probably want to separate I/O errors from other errors.
- I had a very fine-grained `ErrorKind` enum, but this seems to make downstream error handling a bit dreadful. So, I decided to make one enum variant `Format` for file format errors (something different in the format than expected). It's an enum to keep the possibility for more fine-grained errors open.

The `Error` with one variant for crate-specific errors, which wraps `ErrorKind` was inspired by a lot of other crates that follow this approach, where an Error enum wraps the different downstream error types and has one variant for crate-specific errors.

~What I hope to do in a future PR: more fine-grained errors than just an I/O error when reading/writing data, so that we can return errors like `Failed to read embedding matrix, reason: some io::Error explanation`. But I first wanted to have the basic structure set up for errors.~ Added to this PR!

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -5,7 +5,6 @@ use std::io::{BufReader, Read, Seek, SeekFrom, Write};
 use std::mem::size_of;
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
-use failure::{ensure, format_err, Error};
 use memmap::{Mmap, MmapOptions};
 use ndarray::{Array, Array1, Array2, ArrayView, ArrayView2, ArrayViewMut2, Dimension, Ix1, Ix2};
 use rand::{FromEntropy, Rng};
@@ -13,6 +12,7 @@ use rand_xorshift::XorShiftRng;
 use reductive::pq::{QuantizeVector, ReconstructVector, TrainPQ, PQ};
 
 use crate::io::private::{ChunkIdentifier, MmapChunk, ReadChunk, TypeId, WriteChunk};
+use crate::io::{Error, ErrorKind, Result};
 use crate::util::padding;
 
 /// Copy-on-write wrapper for `Array`/`ArrayView`.
@@ -61,11 +61,8 @@ pub struct MmapArray {
 }
 
 impl MmapChunk for MmapArray {
-    fn mmap_chunk(read: &mut BufReader<File>) -> Result<Self, Error> {
-        ensure!(
-            read.read_u32::<LittleEndian>()? == ChunkIdentifier::NdArray as u32,
-            "invalid chunk identifier for NdArray"
-        );
+    fn mmap_chunk(read: &mut BufReader<File>) -> Result<Self> {
+        ChunkIdentifier::ensure_chunk_type(read, ChunkIdentifier::NdArray)?;
 
         // Read and discard chunk length.
         read.read_u64::<LittleEndian>()?;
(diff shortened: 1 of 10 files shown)


src/storage.rs
```diff
@@ -61,40 +61,49 @@ pub struct MmapArray {
...
+        let cols = read.read_u32::<LittleEndian>().map_err(|e| {
+            ErrorKind::io_error("Cannot read number of columns of the embedding matrix", e)
+        })? as usize;
         let shape = Ix2(rows, cols);
 
-        ensure!(
-            read.read_u32::<LittleEndian>()? == f32::type_id(),
-            "Expected single precision floating point matrix for NdArray."
-        );
+        f32::check_read(read)?;
```
sebpuetz: Without context this is not self-explaining. Could be cleared up with some inline comment, that this line verifies the datatype for the embedding matrix.

src/storage.rs
```diff
@@ -161,37 +192,38 @@ impl NdArray {
...
     {
-        let chunk_id = read.read_u32::<LittleEndian>()?;
-        let chunk_id = ChunkIdentifier::try_from(chunk_id)
-            .ok_or_else(|| format_err!("Unknown chunk identifier: {}", chunk_id))?;
-        ensure!(
-            chunk_id == ChunkIdentifier::NdArray,
-            "Cannot read chunk {:?} as NdArray",
-            chunk_id
-        );
+        ChunkIdentifier::check_read(read, ChunkIdentifier::NdArray)?;
```
sebpuetz: Similar comment like above, `check_read` is not self-explanatory for people either not very familiar with the fifu format or with the crate in general.

src/storage.rs
```diff
@@ -161,37 +192,38 @@ impl NdArray {
...
+        })? as usize;
+        let cols = read.read_u32::<LittleEndian>().map_err(|e| {
+            ErrorKind::io_error("Cannot read number of columns of the embedding matrix", e)
+        })? as usize;
 
-        ensure!(
-            read.read_u32::<LittleEndian>()? == f32::type_id(),
-            "Expected single precision floating point matrix for NdArray."
-        );
+        f32::check_read(read)?;
```
sebpuetz: Same as above

src/text.rs
```diff
@@ -4,6 +4,7 @@
 //!
 //! 1. `ReadText`/`WriteText`: word embeddings in text format. In this
 //!    format, each line contains a word followed by its
+
```
sebpuetz: Probably unintentional newline?

src/util.rs
```diff
@@ -25,3 +27,26 @@ pub fn l2_normalize_array(mut v: ArrayViewMut2<f32>) -> Array1<f32> {
 
     norms.into()
 }
+
+pub fn read_number(reader: &mut BufRead, delim: u8) -> Result<usize> {
```
sebpuetz: `read_number` taking a `delim` argument is somewhat counterintuitive. Maybe add rustdocs to clear this up?

src/storage.rs
danieldk: I have changed all of these by `ensure_chunk_type` and `ensure_data_type` and added rustdoc to these methods.

src/util.rs
danieldk: Note that this is only `pub(crate)`, but I can do a separate PR adding rustdoc.

danieldk:
Thanks for the comments!

src/util.rs
sebpuetz: That one always gets me in PRs. I always assume `pub fn` means public interface. For maintainers it's still nice to have, I guess.

danieldk: >That one always gets me in PRs.

Me too. But `pub(crate)`ing the whole module prevents accidental inclusion. Would be nice if GitHub could include automatic annotations (actual visibility).

danieldk: Something like this: https://drewdevault.com/2019/07/08/Announcing-annotations-for-sourcehut.html

One can always hope ;).

Issue #6, opened 2019-06-05
Adding a method for ngram -> idx mappings


sebpuetz:
I was wondering if there are reasons against exposing a method to get tuples of `(ngram, idx)` for both in vocabulary and out of vocabulary words.

Could be nice to inspect similarity of the subword vectors through the python module. Although, we probably can't get all components that make up in vocab words because we precompute theri representations.

danieldk:
Thanks @NianhengWu for adding this!

I guess the next step would be: refactor the existing `subword_indices` method and exposing these methods in Python.

danieldk:
Oh and maybe you could do a PR to append yourself to the list in `CONTRIBUTORS`?

PR #16, opened 2019-07-11
add methods for ngram -> idx mappings

NianhengWu:
#6

### DIFF
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -81,6 +81,7 @@ pub trait SubwordIndices {
     ///
     /// The largest possible bucket exponent is 64.
     fn subword_indices(&self, min_n: usize, max_n: usize, buckets_exp: usize) -> Vec<u64>;
+    fn ngrams_indices(&self, min_n: usize, max_n: usize, buckets_exp: usize) -> Vec<(String, u64)>;
 }
 
 impl SubwordIndices for str {
@@ -106,6 +107,29 @@ impl SubwordIndices for str {
         }
 
         indices
+    }
+
+    fn ngrams_indices(&self, min_n: usize, max_n: usize, buckets_exp: usize) -> Vec<(String, u64)> {
+        assert!(
+            buckets_exp <= 64,
+            "The largest possible buckets exponent is 64."
+        );
+
+        let mask = if buckets_exp == 64 {
+            !0
+        } else {
+            (1 << buckets_exp) - 1
+        };
+        let chars: Vec<_> = self.chars().collect();
+
+        let mut ngrams_indices = Vec::with_capacity((max_n - min_n + 1) * chars.len());
+        for ngram in NGrams::new(&chars, min_n, max_n) {
+            let mut hasher = FnvHasher::default();
+            ngram.hash(&mut hasher);
+            ngrams_indices.push((ngram.into_iter().collect(), hasher.finish() & mask));
+        }
+
+        ngrams_indices
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/subword.rs
```diff
@@ -81,6 +81,7 @@ pub trait SubwordIndices {
     ///
     /// The largest possible bucket exponent is 64.
     fn subword_indices(&self, min_n: usize, max_n: usize, buckets_exp: usize) -> Vec<u64>;
+    fn ngrams_indices(&self, min_n: usize, max_n: usize, buckets_exp: usize) -> Vec<(String, u64)>;
```
danieldk: Add a new line and rustdoc to document this method.

Since we have a reference to the string itself (`&self`), it would be nicer and more efficient (fewer allocations) to return string references: `Vec<(&str, u64)>`

src/subword.rs
```diff
@@ -107,6 +108,29 @@ impl SubwordIndices for str {
 
         indices
     }
+
+    fn ngrams_indices(&self, min_n: usize, max_n: usize, buckets_exp: usize) -> Vec<(String, u64)> {
```
danieldk: This shares most code with `subword_indices`. The overlap should be factored out. Maybe it would be possible to create an iterator that produces `(&str, u64)` iterms. `ngram_indices` could just collect these, `subword_indices` could `map` over the iterator, only retaining the indices. With inlining/optimizations, the actual string slices would probably never be produced by `subword_indices`.

src/vocab.rs
```diff
@@ -153,6 +153,24 @@ impl SubwordVocab {
             Some(indices)
         }
     }
+
+    fn ngrams_indices(&self, word: &str) -> Option<Vec<(String, usize)>> {
```
danieldk: Can also be `Vec<(&str, usize)>`.

Is this made public somewhere? Also, since this is non-essential functionality, but for very specific use cases, we may want to put this method in a separate trait that users could import when they need this functionality.

src/subword.rs
sebpuetz: I don't think `&str` is possible as return type due to the `Vec<char>` being owned by the method. 

I would prefer to get `Vec<(Vec<char>, u64)>` as a return type from this method, since we can get mismatches between the length of `String` vs. `Vec<char>` otherwise and we extract ngrams based on the length of `&[char]`.

danieldk: > I don't think `&str` is possible as return type due to the `Vec<char>` being owned by the method.

`NGrams` was made to work with any data type, but that is not really necessary in this crate. We could specialize n-grams to take `&str` and return an iterator over `&str`. We only have to ensure that hashing is still carried out the same way, but it is possible to iterator over a string character-wise.

sebpuetz: But iirc, slicing `&str`s results in panics if invalid UTF-8 is returned? So just incrementing the beginning of the slice 1-by-1 might lead to crashes, unless I'm missing something.

danieldk: You wouldn't have to increment the beginning one by one. Use `char_indices` to get the offsets of each character in the string and you can make slices that are always valid UTF-8.

danieldk: Though, with all these changes, it would be good to have some benchmarks as well. So, I propose three PRs:

1. Add a benchmark for `NGrams`.
2. Change the `NGrams` implementation to return `&str`. Use (1) to ensure that we are not (much) slower.
3. This PR (rebased on (2)).

sebpuetz: Thanks for the explanation, I wasn't aware of `char_indices`. Is there also a cheap way to get the number of `char`s in a `&str` without collecting `chars()`?

In that case, we'd have to implement `Iterator` for `NGrams<'a, &str>` and `NGram<'a, &[char]>` since specialization isn't stable yet?

danieldk: > Is there also a cheap way to get the number of `char`s in a `&str` without collecting `chars()` (or `char_indices`)?

Nope, you have to decode in linear time due to variable-length encoding.

danieldk: @NianhengWu In `master` `NGrams` returns string slices, so all should be set up to modify this PR to return `Vec<(&str, u64)>`.

NianhengWu: yes i noticed in `master`, `NGrams`  just changed, should I close this PR and start a new one?

danieldk: > yes i noticed in `master`, `NGrams` just changed, should I close this PR and start a new one?

You could rebase this branch on master and incorporate the changes or do a new PR. Whichever you find more convenient ;).

PR #17, opened 2019-07-12
Add a criterion benchmark for the NGrams iterator

danieldk:
I had to make the `subword` module visible (for now), since benchmarks use the library as an external module. I think we can eventually resolve this, by benchmarking Nicole's proposed method.

`cargo bench` will show statistics in the terminal. If gnuplot is installed, nice graphs will also be generated.

Gnuplot

### DIFF
--- a/benches/subword.rs
+++ b/benches/subword.rs
@@ -1,0 +1,33 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use rand::distributions::Alphanumeric;
+use rand::{thread_rng, Rng};
+
+use finalfusion::subword::SubwordIndices;
+
+const MIN_N: usize = 3;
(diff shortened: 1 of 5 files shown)


benches/subword.rs
```diff
@@ -0,0 +1,28 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use rand::distributions::Alphanumeric;
+use rand::{thread_rng, Rng};
+
+use finalfusion::subword::NGrams;
+
+const MIN_N: usize = 3;
+const MAX_N: usize = 3;
```
sebpuetz: Is `MIN_N == MAX_N` intentional?

danieldk: Nope, fixed, thanks!

PR #18, opened 2019-07-12
Specialize NGrams to string slices

danieldk:
NGram::new now takes a string slice, rather than a Vec of a generic
type. The resulting iterator returns the n-grams as string slices.

In SubwordIndices::subword_indices, we used to compute the hash of an
n-gram over Vec<char>. The implementation of Hash for Vec first adds
the length of the Vec to the hasher, followed by the individual
element. However, we want to avoid collecting the characters in a Vec
or iterating over the characters twice in order to get the length.
For this reason, NGrams now couples string slices with their length in
characters using the auxiliary StrWithCharLen type.

---

I have added benchmarks to CI in the previous PR, so we should be able to see the performance relative to the old implementation in the build log for Rust stable.

### DIFF
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -1,7 +1,35 @@
 use std::cmp;
+use std::collections::VecDeque;
 use std::hash::{Hash, Hasher};
+use std::ops::Deref;
 
 use fnv::FnvHasher;
+
+use crate::util::CollectWithCapacity;
+
+/// A string reference with its length in characters.
+struct StrWithCharLen<'a> {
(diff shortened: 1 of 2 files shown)


danieldk:
Criterion says it's about 12% percent faster. I just realized that there may be some more lower hanging fruit in eliminating branches.

danieldk:
The extra branch eliminations only gave a mild (5%) improvement at the cost of introducing `unsafe`, so I'll leave it as-is.

PR #19, opened 2019-07-17
Methods for ngram -> idx mappings added

NianhengWu:
Original issue: #6
Changes:
1. Created a Iterator `NGramsIndicesIter` that produces (&str, u64);
2. Added a trait which has a method for ngram -> idx mappings: `ngrams_indices`;
3. Added relative test module.

### DIFF
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -157,13 +157,84 @@ impl SubwordIndices for str {
     }
 }
 
+/// Iterator over the n-grams in a word and the corresponding subword indices.
+///
+/// `NGramsIndicesIter` is an iterator that produces the n-grams in a word and
+/// the corresponding subword indices as tuples `(ngram, index)`.
+///
+/// **Warning:** no guarantee is provided with regard to the iteration
+/// order. The iterator only guarantees that all n-grams and their indices are produced.
+struct NGramsIndicesIter<'a> {
+    ngrams: NGrams<'a>,
+    mask: u64,
+}
+
+impl<'a> NGramsIndicesIter<'a> {
+    /// Create a new ngrams-indices iterator.
+    ///
+    /// The iterator will create all ngrams of length *[min_n, max_n]* and corresponding
+    /// subword indices.
+    pub fn new(string: &'a str, min_n: usize, max_n: usize, buckets_exp: usize) -> Self {
+        assert!(
+            buckets_exp <= 64,
+            "The largest possible buckets exponent is 64."
+        );
+
+        let mask = if buckets_exp == 64 {
+            !0
+        } else {
+            (1 << buckets_exp) - 1
+        };
+
+        let ngrams = NGrams::new(string, min_n, max_n);
+        NGramsIndicesIter { ngrams, mask }
+    }
+}
+
+impl<'a> Iterator for NGramsIndicesIter<'a> {
+    type Item = (&'a str, u64);
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        self.ngrams.next().map(|ngram| {
+            let mut hasher = FnvHasher::default();
+            ngram.hash(&mut hasher);
+            (ngram.inner, hasher.finish() & self.mask)
+        })
+    }
+}
+
+/// A trait for getting ngrams and their indices of a string.
+///
+/// N-gram indexing assigns an identifier to each subword (n-gram) of a
+/// string. A subword is indexed by computing its hash and then mapping
+/// the hash to a bucket.
+///
+/// Since a non-perfect hash function is used, multiple subwords can
+/// map to the same index.
+pub trait NGramsIndices {
+    /// Return the ngrams and their indices of a string.
+    ///
+    /// The n-grams that are used are of length *[min_n, max_n]*, these are
+    /// mapped to indices into *2^buckets_exp* buckets.
+    ///
+    /// The largest possible bucket exponent is 64.
+    fn ngrams_indices(&self, min_n: usize, max_n: usize, buckets_exp: usize) -> Vec<(&str, u64)>;
+}
+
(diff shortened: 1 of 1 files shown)


NianhengWu:
I think it would better if I open an another PR to change the `subword_indices` method into using the new iterator

src/subword.rs
```diff
@@ -157,13 +157,85 @@ impl SubwordIndices for str {
     }
 }
 
+/// Create an Iterator going through a sentence
```
danieldk: sentence -> token

Maybe something like:

```
Iterator over the n-grams in a word and the corresponding subword indices.

`NGramsIndicesIter` is an iterator that produces the n-grams in a word and
the corresponding subword indices as tuples `(ngram, index)`.
```

I wouldn't mention the types, since they are already added automatically to the rustdoc.

src/subword.rs
```diff
@@ -157,13 +157,85 @@ impl SubwordIndices for str {
...
+///
+/// **Warning:** no guarantee is provided with regard to the iteration
+/// order. The iterator only guarantees that all n-grams and their indices are produced.
+struct NGramsIndicesIter<'a> {
+    ngrams: NGrams<'a>,
+    mask: u64,
+}
+
+impl<'a> NGramsIndicesIter<'a> {
+    pub fn new(string: &'a str, min_n: usize, max_n: usize, buckets_exp: usize) -> Self {
```
danieldk: Add rustdoc to new.

src/subword.rs
```diff
@@ -157,13 +157,85 @@ impl SubwordIndices for str {
...
+        let mask = if buckets_exp == 64 {
+            !0
+        } else {
+            (1 << buckets_exp) - 1
+        };
+
+        let ngrams = NGrams::new(string, min_n, max_n);
+        NGramsIndicesIter { ngrams, mask }
+    }
+}
```
danieldk: Put a newline between `impl` blocks.

src/subword.rs
```diff
@@ -157,13 +157,85 @@ impl SubwordIndices for str {
...
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        let next_ngram = self.ngrams.next();
+        match next_ngram {
+            Some(ngram) => {
+                let mut hasher = FnvHasher::default();
+                ngram.hash(&mut hasher);
+                Some((ngram.inner, hasher.finish() & self.mask))
+            }
+            None => return None,
```
danieldk: Superfluous return, since this is the last expression of `next`, this can be

```
None => None,
```

However, since you only have to do something in the `Some` case, the function body can be simplified to:

```rust
self.ngrams.next().map(|ngram| {
  let mut hasher = FnvHasher::default();
  ngram.hash(&mut hasher);
  (ngram.inner, hasher.finish() & self.mask)
}
```

src/subword.rs
```diff
@@ -157,13 +157,85 @@ impl SubwordIndices for str {
...
+        }
+    }
+}
+
+/// A trait for getting ngrams and their indices in a sentence.
```
danieldk: `in a sentence.` -> `of a string.`

src/subword.rs
```diff
@@ -250,6 +322,19 @@ mod tests {
         };
     }
 
+    lazy_static! {
+        static ref NGRAMS_INDICES_TESTS_36: HashMap<&'static str, Vec<(&'static str, u64)>> = hashmap! {
+            "<Daniël>" =>
+                vec![("Dan",214157), ("iël",233912), ("Danië",311961), ("iël>",488897), ("niël>",620206), ("anië",741276), ("Dani",841219),
```
danieldk: Run `cargo fmt`.

It seems that CI doesn't fail with incorrect formatting, though it should. I'll fix that.

NianhengWu: Sorry, I'm not so sure which part of it is incorrect? I ran `cargo fmt` and `cargo fmt --all -- --check` on it and it turns out nothing's changed.

danieldk: Tuple elements have to be separated by a space (e.g. `("Dan", 214157)`). Also the lines are too long. `rustfmt` does not seem to fix these cases because they are in a structure-matching macro call.

danieldk: You could hand-format it, or replace it by something that `rustfmt` can format, e.g.:

```rust
    lazy_static! {
        static ref NGRAMS_INDICES_TESTS_36: HashMap<&'static str, Vec<(&'static str, u64)>> = [
            (
                "<Daniël>",
                vec![
                    ("Dan", 214157),
                    ("iël", 233912),
                    ("Danië", 311961),
                    ("iël>", 488897),
                    ("niël>", 620206),
                    ("anië", 741276),
                    ("Dani", 841219),
                    ("Daniël", 1167494),
                    ("ani", 1192256),
                    ("niël", 1489905),
                    ("ël>", 1532271),
                    ("nië", 1644730),
                    ("<Dan", 1666166),
                    ("aniël", 1679745),
                    ("<Danië", 1680294),
                    ("aniël>", 1693100),
                    ("<Da", 2026735),
                    ("<Dani", 2065822)
                ]
            ),
            (
                "<hallo>",
                vec![
                    ("lo>", 75867),
                    ("<hal", 104120),
                    ("hallo>", 136555),
                    ("hal", 456131),
                    ("allo>", 599360),
                    ("llo", 722393),
                    ("all", 938007),
                    ("<ha", 985859),
                    ("hallo", 1006102),
                    ("allo", 1163391),
                    ("llo>", 1218704),
                    ("<hallo", 1321513),
                    ("<hall", 1505861),
                    ("hall", 1892376)
                ]
            )
        ]
        .iter()
        .cloned()
        .collect();
    }
```
```

NianhengWu: Wow, thank you!!

PR #23, opened 2019-07-24
Add Similarity query for embeddings.

sebpuetz:
Add a method to query for the most similar words given a query vector.

Thought, this could be nice to have since analogy queries are essentially doing the same.

### DIFF
--- a/src/similarity.rs
+++ b/src/similarity.rs
@@ -16,12 +16,12 @@ use crate::vocab::Vocab;
 /// This data structure is used to store a pair consisting of a word and
 /// its similarity to a query word.
 #[derive(Debug, Eq, PartialEq)]
-pub struct WordSimilarity<'a> {
+pub struct WordSimilarityResult<'a> {
     pub similarity: NotNan<f32>,
     pub word: &'a str,
 }
 
-impl<'a> Ord for WordSimilarity<'a> {
+impl<'a> Ord for WordSimilarityResult<'a> {
     fn cmp(&self, other: &Self) -> Ordering {
         match other.similarity.cmp(&self.similarity) {
             Ordering::Equal => self.word.cmp(other.word),
@@ -30,8 +30,8 @@ impl<'a> Ord for WordSimilarity<'a> {
     }
 }
 
-impl<'a> PartialOrd for WordSimilarity<'a> {
-    fn partial_cmp(&self, other: &WordSimilarity) -> Option<Ordering> {
+impl<'a> PartialOrd for WordSimilarityResult<'a> {
+    fn partial_cmp(&self, other: &WordSimilarityResult) -> Option<Ordering> {
         Some(self.cmp(other))
     }
 }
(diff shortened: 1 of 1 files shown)


sebpuetz:
@danieldk 
I don't have write access to this repo, so I can't request a review.

danieldk:
@sebpuetz I have given you owner privileges for the whole finalfusion organization.

sebpuetz:
> 
> 
> @sebpuetz I have given you owner privileges for the whole finalfusion organization now.

Oh nice, thanks!

src/similarity.rs
```diff
@@ -194,16 +194,36 @@ pub trait Similarity {
...
-    fn similarity(&self, word: &str, limit: usize) -> Option<Vec<WordSimilarity>>;
+    fn word_similarity(&self, word: &str, limit: usize) -> Option<Vec<WordSimilarity>>;
+
+    /// Find words that are similar to the query embedding.
+    ///
+    /// The similarity between the query embedding and other embeddings is
+    /// defined by the dot product of the embeddings. If the vectors are unit
+    /// vectors (e.g. by virtue of calling `normalize`), this is the cosine
+    /// similarity. At most, `limit` results are returned.
+    fn embedding_similarity(
```
danieldk: I don't like overpacking traits with somewhat different functionality. Could you split this into two traits: `WordSimilarity` and `EmbeddingSimilarity`. We could consider folding `word_similarity_by` and `embedding_similarity_by` into these two respective traits.

src/similarity.rs
```diff
@@ -445,13 +493,31 @@ mod tests {
...
+        let f = File::open("testdata/similarity.bin").unwrap();
+        let mut reader = BufReader::new(f);
+        let embeddings = Embeddings::read_word2vec_binary(&mut reader).unwrap();
+        let embedding = embeddings.embedding("Berlin").unwrap();
+        let result = embeddings.embedding_similarity(embedding.as_view(), 10);
+        assert!(result.is_some());
+        let mut result = result.unwrap().into_iter();
+        assert_eq!(10, result.len());
+        assert_eq!(result.next().unwrap().word, "Berlin");
+        println!("{:?}", result);
```
danieldk: Stray `println!`. I know it's also in the other test, but we shouldn't introduce it in this test as well ;).

src/similarity.rs
sebpuetz: Did that, although I had to rename the `WordSimilarity` struct to `WordSimilarityResult`

sebpuetz:
> This is a nice addition! People can probe with their custom vectors.

What do you think of adding a mask to these queries? Take a `HashSet<&str>` to allow skipping certain neighbours, rather than defaulting to an empty `HashSet`?

sebpuetz:
Seems like my branch was coming from my fork's master branch which wasn't up to date. I'm rebasing on finalfusion/finalfusion-rust master now.

sebpuetz:
The PR should no be mergeable without conflicts

danieldk:
Great! Feel free to merge it if CI doesn't complain.

PR #24, opened 2019-07-26
Use arrays for analogy queries.

sebpuetz:
Replace `word1: &str, word2: &str, word3: &str` with `[&str; 3]` in analogy queries

### DIFF
--- a/src/similarity.rs
+++ b/src/similarity.rs
@@ -51,12 +51,10 @@ pub trait Analogy {
     /// indicating which of the tokens were present.
     fn analogy(
         &self,
-        word1: &str,
-        word2: &str,
-        word3: &str,
+        query: [&str; 3],
         limit: usize,
     ) -> Result<Vec<WordSimilarityResult>, [bool; 3]> {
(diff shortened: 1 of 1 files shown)


src/similarity.rs
```diff
@@ -78,9 +76,7 @@ pub trait Analogy {
     /// were present.
     fn analogy_masked(
         &self,
-        word1: &str,
-        word2: &str,
-        word3: &str,
+        query: [&str; 3],
         limit: usize,
         remove: [bool; 3],
```
danieldk: I think it would be best to make `limit` the last argument, so that all word-related arguments are together.

src/similarity.rs
```diff
@@ -93,16 +89,12 @@ where
 {
     fn analogy_masked(
         &self,
-        word1: &str,
-        word2: &str,
-        word3: &str,
+        query: [&str; 3],
         limit: usize,
         remove: [bool; 3],
```
danieldk: Same here.

sebpuetz:
Sure, makes sense. I moved `limit` to the end of the arguments.

PR #25, opened 2019-07-31
Separate index computation from subword iteration

danieldk:
The extraction of subwords + subword indices was tied to one specific
implementation of index generation (finalfusion-style FNV hashing +
bucket mapping). This change separates index generation to an
`Indexer` trait + `HashIndexer` implementation.

This change accomodates future support of other indexing functions
(such as fastText).

### DIFF
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -1,14 +1,99 @@
 use std::cmp;
 use std::collections::VecDeque;
+use std::fmt;
 use std::hash::{Hash, Hasher};
+use std::marker::PhantomData;
 use std::ops::Deref;
 
 use fnv::FnvHasher;
 
 use crate::util::CollectWithCapacity;
 
+/// N-Gram indexer
+///
+/// An indexer maps an n-gram to an index in the subword embedding
+/// matrix.
+pub trait Indexer {
+    /// Map an n-gram to an index in the subword embedding matrix.
+    fn index_ngram(&self, ngram: &StrWithCharLen) -> u64;
+}
+
+/// Indexer using a hash function.
+///
+/// This indexer first hashes a given n-gram and then maps the
+/// resulting hash into *2^buckets_exp* buckets.
+///
+/// The largest possible bucket exponent is 64.
+pub struct HashIndexer<H> {
+    mask: u64,
+    _phantom: PhantomData<H>,
+}
+
+impl<H> HashIndexer<H> {
+    /// Construct a `HashIndexer`.
+    ///
+    /// The largest possible bucket exponent is 64.
+    pub fn new(buckets_exp: usize) -> Self {
+        assert!(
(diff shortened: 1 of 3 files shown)


src/subword.rs
```diff
@@ -1,14 +1,99 @@
...
+/// resulting hash into *2^buckets_exp* buckets.
+///
+/// The largest possible bucket exponent is 64.
+pub struct HashIndexer<H> {
+    mask: u64,
+    _phantom: PhantomData<H>,
+}
+
+impl<H> HashIndexer<H> {
+    /// Construct an `FNV1Indexer`.
```
sebpuetz: Maybe a typo?
```Rust
 /// Construct a `HashIndexer`
```

src/subword.rs
```diff
@@ -1,14 +1,99 @@
...
+    }
+}
+
+impl<H> Eq for HashIndexer<H> {}
+
+impl<H> Indexer for HashIndexer<H>
+where
+    H: Default + Hasher,
+{
+    fn index(&self, ngram: &StrWithCharLen) -> u64 {
```
sebpuetz: There's a collision with `Index::index`, I'm not sure whether this could lead to any confusion

src/subword.rs
```diff
@@ -125,15 +210,20 @@ pub trait SubwordIndices {
     /// Return the subword indices of the subwords of a string.
     ///
     /// The n-grams that are used are of length *[min_n, max_n]*, these are
-    /// mapped to indices into *2^buckets_exp* buckets.
+    /// mapped to indices using the given indexed.
```
sebpuetz: *`the given indexer.`

src/subword.rs
sebpuetz: Or this could be implemented through `Index::index`. In case of an actual ngram lookup, that would even be more fitting. In case of only using the bucket it's also not too far fetched.

danieldk: > There's a collision with `Index::index`, I'm not sure whether this could lead to any confusion

I think normally it is not a problem, but since is so prominent, I renamed the method to `index_ngram` just to be certain.

danieldk: > Or this could be implemented through `Index::index`. In case of an actual ngram lookup, that would even be more fitting. In case of only using the bucket it's also not too far fetched.

That would be very elegant, but the problem is that `Index::index` is required to return a reference:

https://doc.rust-lang.org/std/ops/trait.Index.html#tymethod.index

which we can't :(.

danieldk:
Thanks for the review and suggested changes!

src/subword.rs
sebpuetz: > That would be very elegant, but the problem is that `Index::index` is required to return a reference:

Ah, I forgot about that. Unfortunate!

PR #32, opened 2019-08-02
Add and implement the ReadFastText trait

danieldk:
I guess that the build will fail, because new `serde` versions had incompatibilities with Rust 1.31. If so, I'll try to downgrade the dependency.

### DIFF
--- a/src/fasttext/io.rs
+++ b/src/fasttext/io.rs
@@ -1,0 +1,327 @@
+use std::io::BufRead;
+
+use byteorder::{LittleEndian, ReadBytesExt};
+use ndarray::{s, Array2, ErrorKind as ShapeErrorKind, ShapeError};
+use serde::Serialize;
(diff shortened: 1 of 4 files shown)


src/fasttext/io.rs
```diff
@@ -0,0 +1,327 @@
...
+    if prune_idx_size > 0 {
+        return Err(ErrorKind::Format("Pruned vocabularies are not supported".into()).into());
+    }
+
+    let mut words = Vec::with_capacity(size as usize);
+    for _ in 0..size {
+        let word = read_string(reader, 0)?;
+        reader
+            .read_u64::<LittleEndian>()
+            .map_err(|e| ErrorKind::io_error("Cannot word frequency", e))?;
```
sebpuetz: *read

danieldk: Fixed

Issue #38, opened 2019-08-05
Delimiters and trimming in read_text_dims


sebpuetz:
I was converting some files from the text(dims) format to our format and ran into trouble with tokens containing whitespaces. There was a number of tokens with non-space whitespaces within tokens which threw our splitting through `.split_whitespace()` off. 

Also, there were a bunch of tokens that were just a whitespace (e.g. NEL `U+0085` and NBSP `U+00A0`), getting reduced to empty strings by the `.trim()` call we do on the words. The second part was particularly frustrating to debug, since `words` ended up with the correct length, the right number of embeddings was created but an incorrect number of vocabulary entries were being created. Which happened because `trim()`ming reduced a couple hundred tokens to `""`.

https://github.com/finalfusion/finalfusion-rust/blob/c9bd67c88c3b4e46bd7a3775271e8d0244e5b643/src/compat/text.rs#L161-L170

This took an awfully long time to debug, mostly due to the stupid format and the unnormalized corpus that the embeddings were trained on, but still, I wonder if we can do better here. E.g.:

```Rust
        let word = parts
            .next()
            .ok_or_else(|| ErrorKind::Format(String::from("Spurious empty line")))?
            .trim_matches(|c| c == ' ' || c == '\t' );
```

What do you think about targeted trimming and splitting on `'\t'` and `'  '` rather than any whitespaces? Might reduce the headaches of potential users.

______

While debugging this, I also found that the following assertion leads to killing the Python process when loading a malformed file through our Python library. https://github.com/finalfusion/finalfusion-rust/blob/c9bd67c88c3b4e46bd7a3775271e8d0244e5b643/src/embeddings.rs#L50-L54

It's kind of ugly to get this kind of output for a malformed file, There's an issue at https://github.com/PyO3/pyo3/issues/492 about stopping panics at the ffi boundary, but for now we would have to deal with that ourselves. Maybe it's possible to return a `Result` here?

```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `199872`,
 right: `200000`: Vocab and norms do not have the same length', /data/rust_projects/finalfusion-rust/src/embeddings.rs:50:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
fatal runtime error: failed to initiate panic, error 5
[1]    8362 abort (core dumped)  ipython
```

danieldk:
Great debugging work!

[...]

PR #39, opened 2019-08-05
Change splitting and trimming for textdims.

sebpuetz:
Add more targeted splitting and trimming by only trimming and
splitting on tabs and spaces rather than any unicode whitespace.

#38

### DIFF
--- a/src/compat/text.rs
+++ b/src/compat/text.rs
@@ -161,12 +161,14 @@ where
     for line in reader.lines() {
         let line =
             line.map_err(|e| ErrorKind::io_error("Cannot read line from embedding file", e))?;
-        let mut parts = line.split_whitespace();
+        let mut parts = line
+            .split(|c: char| c.is_ascii_whitespace())
+            .filter(|part| !part.is_empty());
 
         let word = parts
             .next()
             .ok_or_else(|| ErrorKind::Format(String::from("Spurious empty line")))?
-            .trim();
+            .trim_matches(|c: char| c.is_ascii_whitespace());
         words.push(word.to_owned());
 
         for part in parts {


src/compat/text.rs
```diff
@@ -161,12 +161,14 @@ where
     for line in reader.lines() {
         let line =
             line.map_err(|e| ErrorKind::io_error("Cannot read line from embedding file", e))?;
-        let mut parts = line.split_whitespace();
+        let mut parts = line
+            .split(|c| c == ' ' || c == '\t')
+            .filter(|c| !c.is_empty());
```
danieldk: Maybe rename `c` here, because the elements are not characters anymore?

Also, maybe you could put a comment here that once we have 1.34.0 as a minimum, this can be replaced by `split_ascii_whitespace`?

https://doc.rust-lang.org/std/primitive.str.html#method.split_ascii_whitespace

src/compat/text.rs
```diff
@@ -161,12 +161,14 @@ where
     for line in reader.lines() {
         let line =
             line.map_err(|e| ErrorKind::io_error("Cannot read line from embedding file", e))?;
-        let mut parts = line.split_whitespace();
+        let mut parts = line
+            .split(|c| c == ' ' || c == '\t')
```
danieldk: maybe `.split(char::is_ascii_whitespace)` ?

src/compat/text.rs
```diff
@@ -161,12 +161,14 @@ where
...
+            .filter(|c| !c.is_empty());
 
         let word = parts
             .next()
             .ok_or_else(|| ErrorKind::Format(String::from("Spurious empty line")))?
-            .trim();
+            .trim_matches(|c| c == ' ' || c == '\t');
```
danieldk: Would

`.trim_matches(char::is_ascii_whitespace)`

work?

src/compat/text.rs
sebpuetz: Unfortunately `is_ascii_whitespace` takes a `&char`, we get `char` here. So compiling fails with:

```
    |
165 |             .split(char::is_ascii_whitespace)
    |              ^^^^^
    |              |
    |              expected signature of `fn(char) -> _`
    |              found signature of `for<'r> fn(&'r char) -> _`
    |
```

src/compat/text.rs
sebpuetz: > Maybe rename c here, because the elements are not characters anymore?

Yeah, `part` makes more sense.

> Also, maybe you could put a comment here that once we have 1.34.0 as a minimum, this can be replaced by split_ascii_whitespace?

Weird, that works. I was pretty sure that I tried that yesterday. Probably hadn't yet realized that I was trimming words into nothingness.

src/compat/text.rs
sebpuetz: Pushed out the changes, trimming and splitting now happens based on `is_ascii_whitespace`.

sebpuetz:
Pushed out the changes, trimming and splitting now happens based on is_ascii_whitespace.

danieldk:
Maybe this change should be cherry-picked on the 0.8.0 branch as well?

sebpuetz:
> Maybe this change should be cherry-picked on the 0.8.0 branch as well?

Sure, would definitely improve usability.

Although one more thing I'm wondering about: Should we return an `Err` if a `word` is empty? I don't think associating the empty String with an embedding makes any sense.

danieldk:
> > Maybe this change should be cherry-picked on the 0.8.0 branch as well?
> 
> Sure, would definitely improve usability.
> 
> Although one more thing I'm wondering about: Should we return an `Err` if a `word` is empty? I don't think associating the empty String with an embedding makes any sense.

But wouldn't it be filtered out? And then the number of vector components wouldn't be correct anymore.

sebpuetz:
> > > Maybe this change should be cherry-picked on the 0.8.0 branch as well?
> > 
> > 
> > Sure, would definitely improve usability.
> > Although one more thing I'm wondering about: Should we return an `Err` if a `word` is empty? I don't think associating the empty String with an embedding makes any sense.
> 
> But wouldn't it be filtered out? And then the number of vector components wouldn't be correct anymore.

Not if this makes more than one word the empty string:

```Rust
        let word = parts
            .next()
            .ok_or_else(|| ErrorKind::Format(String::from("Spurious empty line")))?
            .trim_matches(|c: char| c.is_ascii_whitespace());
```

Then the assertion that I mentioned in the issue will fail, leaving you wonder why there is the right number of norms but not enough vocabulary entries. Although with the changes from this PR, it should be much harder to run into such issues.

danieldk:
> ```rust
>         let word = parts
>             .next()
>             .ok_or_else(|| ErrorKind::Format(String::from("Spurious empty line")))?
>             .trim_matches(|c: char| c.is_ascii_whitespace());
> ```
> 
> Then the assertion that I mentioned in the issue will fail, leaving you wonder why there is the right number of norms but not enough vocabulary entries. Although with the changes from this PR, it should be much harder to run into such issues.

I was looking at the wrong code. I think it indeed makes sense to put in an empty word check then.

PR #40, opened 2019-08-07
Add read(_dims)?_lossy

danieldk:
The lossy read variants allow invalid UTF-8 byte sequences in tokens.
Such sequences are replaced by the replacement character.

Will do the lossy variant of word2vec in a separate PR.

### DIFF
--- a/src/compat/text.rs
+++ b/src/compat/text.rs
@@ -56,6 +56,14 @@ where
 {
     /// Read the embeddings from the given buffered reader.
     fn read_text(reader: &mut R) -> Result<Self>;
+
+    /// Read the embeddings from the given buffered reader.
+    ///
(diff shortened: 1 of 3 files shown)


danieldk:
I have changed my mind and will make lossy reading a `bool` argument. This differs from `String::from_utf8_lossy` in that that requires a different type signature.

danieldk:
> Would be neat if there was a `lines_lossy` method, but it seems like that doesn't exist.

Indeed, would be fairly trivial to implement. Or maybe a lines_bytes that would return lines as `Vec<u8>` or perhaps `Cow<'a, [u8]>` so that it could share the underlying buffer if possible. Then the user could decide what to do with the bytes.

PR #43, opened 2019-08-12
Add read_fasttext_lossy

danieldk:


danieldk:
I'll cherry pick all these lossy read methods on the latest release branch. Assuming that nobody is implementing these traits outside the crate (they shouldn't ;)), these do not entail API changes.

PR #45, opened 2019-08-29
Verify that the vocabulary indexing matches the embeddings shape.

sebpuetz:
Without the assertion, it is possible to construct `Embeddings` where the vocabulary can return out of bounds indices.

### DIFF
--- a/src/chunks/vocab.rs
+++ b/src/chunks/vocab.rs
@@ -417,8 +417,11 @@ pub trait Vocab: Clone {
     /// Get the index of a token.
     fn idx(&self, word: &str) -> Option<WordIndex>;
 
-    /// Get the vocabulary size.
-    fn len(&self) -> usize;
+    /// Get the number of words in the vocabulary.
+    fn words_len(&self) -> usize;
+
+    /// Get the total length of this vocabulary, including possible subword indices.
+    fn vocab_len(&self) -> usize;
 
     /// Get the words in the vocabulary.
     fn words(&self) -> &[String];
@@ -429,8 +432,12 @@ impl Vocab for SimpleVocab {
         self.indices.get(word).cloned().map(WordIndex::Word)
     }
 
-    fn len(&self) -> usize {
+    fn words_len(&self) -> usize {
         self.indices.len()
+    }
+
+    fn vocab_len(&self) -> usize {
+        self.words_len()
     }
 
     fn words(&self) -> &[String] {
@@ -452,8 +459,12 @@ where
         self.subword_indices(word).map(WordIndex::Subword)
     }
 
-    fn len(&self) -> usize {
+    fn words_len(&self) -> usize {
         self.indices.len()
+    }
+
+    fn vocab_len(&self) -> usize {
+        self.words_len() + self.indexer.upper_bound() as usize
     }
 
     fn words(&self) -> &[String] {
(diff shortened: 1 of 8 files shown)


sebpuetz:
> I am not sure about `len`/`max_idx`. `len` was already very non-descriptive, and one really has to read the documentation and understand the vocabs to understand the difference between `len` and `max_idx`

I agree that we should find a better names for this, but I don't see a way to squash the whole semantics into a single method name. I don't think without knowing about different vocab types (or loking at the docs) it'd be possible to make this transparent.

> And actually, the maximum index is `max_idx` - 1?

Weirdly enough, tests fail with `max_idx-1`.

danieldk:
> I agree that we should find a better names for this, but I don't see a way to squash the whole semantics into a single method name. I don't think without knowing about different vocab types (or loking at the docs) it'd be possible to make this transparent.

Well, it could be something like `words_len` and `vocab_len`. At least there is some parallelism and `len` is actually what it is, not index (see below).

> Weirdly enough, tests fail with `max_idx-1`.

No, I mean if you actually wanted to get the last possible index. The length is always the last possible index plus one. That's one reason the method name is a misnomer.

sebpuetz:
> > I agree that we should find a better names for this, but I don't see a way to squash the whole semantics into a single method name. I don't think without knowing about different vocab types (or loking at the docs) it'd be possible to make this transparent.
> 
> Well, it could be something like `words_len` and `vocab_len`. At least there is some parallelism and `len` is actually what it is, not index (see below).

Then `vocab_len` would be `max_idx` + 1? And `words_len` would replace the current `len` method?

> > Weirdly enough, tests fail with `max_idx-1`.
> 
> No, I mean if you actually wanted to get the last possible index. The length is always the last possible index plus one. That's one reason the method name is a misnomer.

Oops, that makes sense.

sebpuetz:
I think for the `Indexer` it makes sense to return a `max_idx` rather than a `len`, at least the concept of the length of an indexer seems weird.

danieldk:
> I think for the `Indexer` it makes sense to return a `max_idx` rather than a `len`, at least the concept of the length of an indexer seems weird.

Then I would favor `upper_bound` (which is the maximum possible index + 1), because also here for `max_idx` I would expect `len - 1`.

sebpuetz:
I went with `words_len`, `vocab_len` and `upper_bound`, that should convey the meaning properly.

src/compat/fasttext/indexer.rs
```diff
@@ -44,6 +44,10 @@ impl Indexer for FastTextIndexer {
     fn index_ngram(&self, ngram: &StrWithCharLen) -> u64 {
         u64::from(fasttext_hash(ngram.as_str()) % self.buckets)
     }
+
+    fn upper_bound(&self) -> u64 {
+        u64::from(self.buckets) + 1
```
danieldk: `u64::from(self.buckets)` is correct, since the buckets are numbered from 0, the upper bound is `u64::from(self.buckets)`, the largest possible index `u64::from(self.buckets) - 1`

src/subword.rs
```diff
@@ -108,6 +111,11 @@ where
         ngram.hash(&mut hasher);
         hasher.finish() & self.mask
     }
+
+    fn upper_bound(&self) -> u64 {
+        // max val is <= 64
+        2u64.pow(self.buckets_exp as u32) + 1
```
danieldk: `2u64.pow(self.buckets_exp as u32)`

src/chunks/vocab.rs
```diff
@@ -452,10 +459,14 @@ where
         self.subword_indices(word).map(WordIndex::Subword)
     }
 
-    fn len(&self) -> usize {
+    fn words_len(&self) -> usize {
         self.indices.len()
     }
 
+    fn vocab_len(&self) -> usize {
+        self.words_len() + self.indexer.upper_bound() as usize - 1
```
danieldk: `self.words_len() + self.indexer.upper_bound() as usize`

sebpuetz:
Figuring out lengths turns out to be harder than expected. Thanks for the fixes!

danieldk:
Let's merge after CI is happy.

PR #47, opened 2019-08-29
Split the storage module into several submodules

danieldk:
The storage module is becoming large and unwieldy. Split it into several submodules.


sebpuetz:
Should be good to go if CI is happy

PR #48, opened 2019-08-29
Add MmapQuantizedArray

danieldk:
Does what it says on the tin: memory-maps the quantized embeddings
matrix and norms (if present).

This change does not implement WriteChunk for this data type, which
will be added through a separate PR.

---

There is still a build error, because `tempfile` depends on some crate that does not build with Rust 1.31.0. This 'only support/check against the latest Rust stable' in the Rust ecosystem is becoming quite maddening.


sebpuetz:
To merge this, we'd have to drop the 1.31 compatibility, right? 

`tempfile` writes in their git repo that the minimum required Rust version is 1.32, which is also what they are testing against

I'll get around to review this PR tomorrow.

danieldk:
> To merge this, we'd have to drop the 1.31 compatibility, right?

Yes. I liked having 1.31 as the lower bound, since it is the first version of Rust that supports Edition 2018. The editions are nice for supporting incompatible changes, but only solve half of the problem, we call it Rust 2018 (`edition = 2018`), but in practice it means Rust 2018 plus some arbitrary language features that were added after 1.31.0.

I would like to avoid increasing the lower bound until NixOS/nixpkgs 19.09. I am building sticker Docker images against NixOS 19.03, which has Rust 1.31. Although one can trivially use different Rust versions [1], using the nixpkgs stable version has the benefit that a lot of core crates are in the Nix binary cache, so you get very short build times, etc.

It's strange that the functionality of tempfile is not in the standard library.

[1] Which I do for e.g. pyo3: https://git.sr.ht/~danieldk/nix-packages/tree/master/pkgs/python-modules/finalfusion/default.nix#L12

danieldk:
Relevant discussion:

https://github.com/Stebalien/tempfile/issues/100

I propose explicitly setting `tempfile` to < 3.1.0 for now. Then once NixOS/nixpkgs 19.09 is released, we can bump bump the minimum version to Rust 1.32.0.

sebpuetz:
If I see this correctly, the `tempfile` dependency is only for a test case? Is `Cursor` not `mmap`pable?

danieldk:
> If I see this correctly, the `tempfile` dependency is only for a test case? Is `Cursor` not `mmap`pable?

No, you can only memory map files (or anymous memory, but that's more of a low-level virtual memory allocation primitive in UNIX, besides `brk`/`sbrk`).

sebpuetz:
Unfortunate. Creating a non-temp file and deleting that from within the test would be too inelegant?

danieldk:
> Unfortunate. Creating a non-temp file and deleting that from within the test would be too inelegant?

Too many edge cases.

danieldk:
> > Unfortunate. Creating a non-temp file and deleting that from within the test would be too inelegant?
> 
> Too many edge cases. I think the most appropriate alternative would be to include the generated file (as we include some other embedding files now).

Updated with this alternative. Does that look ok?

sebpuetz:
That looks good to me and I think it's nice to avoid additional dependencies. Especially if we'd have to adjust what we support for a test-dependency.

Although I'm wondering what a normal time-frame to up the requirements would be.

danieldk:
> Although I'm wondering what a normal time-frame to up the requirements would be.

Probably soon ;). I was updating `reductive` dependencies and `rand` 0.7 also has some dependency that requires Rust 1.32.0.

PR #50, opened 2019-09-04
Implement WriteChunk for MmapQuantizedArray

danieldk:
The diff is really ugly due to factoring out the chunk writing code. `git diff --histogram` looks much better.

### DIFF
--- a/src/chunks/storage/quantized.rs
+++ b/src/chunks/storage/quantized.rs
@@ -100,6 +100,139 @@ impl QuantizedArray {
             read_norms,
         })
     }
+
+    fn write_chunk<W>(
+        write: &mut W,
+        quantizer: &PQ<f32>,
+        quantized: ArrayView2<u8>,
+        norms: Option<ArrayView1<f32>>,
+    ) -> Result<()>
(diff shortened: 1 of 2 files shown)


sebpuetz:
> I assume this returns the quantized embeddings? The fn name isn't entirely transparent to me, although it's private anyways.

Seems like I commented on the commit instead of the PR. Anyways: `MmapQuantizedArray::quantized` isn't very descriptive, I feel `quantized_embeddings` or `quantized_matrix` or something like that would be clearer.

danieldk:
> Seems like I commented on the commit instead of the PR. Anyways: `MmapQuantizedArray::quantized` isn't very descriptive, I feel `quantized_embeddings` or `quantized_matrix` or something like that would be clearer.

I used the same name as the member variable, since it returns a ndarray view version of `self.quantized`.

danieldk:
> I used the same name as the member variable, since it returns a ndarray view version of `self.quantized`.

I will merge it now and make a separate commit to rename both the member variables and methods to `quantized_array` or something along those lines.

PR #54, opened 2019-09-13
Update to reductive 0.3

danieldk:


danieldk:
Sadly, this is an API change, so we cannot cherry-pick this into `release-0.9`.

sebpuetz:
Considering we don't have much else in the pipeline for `finalfusion`, we could think about an early `0.10` release?

danieldk:
> ~Considering we don't have much else in the pipeline for `finalfusion-rust`, we could think about an early `0.10` release?~
> 
> Nevermind, forgot about pruning.

But I think pruning will take a while to settle, since there experiments need to be done, to see which variation is the best. What is bugging me is that the `finalfusion-utils` that we currently distribute advertises multiple quantization attempts, but never actually does that (instead just burning CPU for trying the same seed centroids over and over again). So, I am in favor of doing a 0.10 release pretty fast. Some optimizations + more flexible ndarray-linalg are cherries on top.

src/chunks/storage/quantized.rs
```diff
@@ -80,23 +80,23 @@ impl QuantizedArray {
...
-                (n_centroids, reconstructed_len / quantized_len),
-                subquantizer_vec,
-            )
-            .map_err(Error::Shape)?;
-            quantizers.push(subquantizer);
-        }
+        let quantizer_shape = (
+            quantized_len,
+            n_centroids,
+            reconstructed_len / quantized_len,
```
sebpuetz: Is there anything to verify that `reconstructed_len % quantized_len == 0`?

danieldk:
First time I bump into this one:

```
error: large size difference between variants

  --> src/chunks/storage/wrappers.rs:25:5

   |

25 |     QuantizedArray(QuantizedArray),

   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

   |

   = note: `-D clippy::large-enum-variant` implied by `-D warnings`

   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant

help: consider boxing the large fields to reduce the total size of the enum

   |

25 |     QuantizedArray(Box<QuantizedArray>),

   |             
```

Nothing fishy seems to be going on, `Array` just has a relatively large footprint (by storing dimensions, strides, and an extra pointer besides the wrapped `Vec`). I will box it.

src/chunks/storage/quantized.rs
danieldk: `reductive` has quite some assertions. This one is not relevant to reductive outside training, since reductive just gets a tensor, that determines the quantizer parameters and the PQ constructor verifies that the projection matrix has the corresponding size.

But there could be malformed files, so we don't want to abort with a failed assertion, but instead return `Err`. I'll make an issue and do a separate PR.

sebpuetz:
> First time I bump into this one:
> 
> ```
> error: large size difference between variants
> 
>   --> src/chunks/storage/wrappers.rs:25:5
> 
>    |
> 
> 25 |     QuantizedArray(QuantizedArray),
> 
>    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
>    |
> 
>    = note: `-D clippy::large-enum-variant` implied by `-D warnings`
> 
>    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
> 
> help: consider boxing the large fields to reduce the total size of the enum
> 
>    |
> 
> 25 |     QuantizedArray(Box<QuantizedArray>),
> 
>    |             
> ```
> 
> Nothing fishy seems to be going on, `Array` just has a relatively large footprint (by storing dimensions, strides, and an extra pointer besides the wrapped `Vec`). I will box it.

First time I see that, too. Looks like the threshold can be configured? https://rust-lang.github.io/rust-clippy/master/#large_enum_variant

src/chunks/storage/quantized.rs
sebpuetz: Given that's really an edge-case, but since we're returning `Result` here anyways...

danieldk: Cosmic ray that bit flips memory while reading/writing the embeddings? ;)

At any rate, I have opened #55 for this. Good catch.

danieldk:
> First time I see that, too. Looks like the threshold can be configured? https://rust-lang.github.io/rust-clippy/master/#large_enum_variant

Indeed. But I want to believe that the 200 is picked for good reasons :).

danieldk:
BTW, I did some benchmarking, no noticeable difference.

PR #58, opened 2019-09-27
Support alignment of embedding matrices on 16-byte boundaries

danieldk:
I spent a fair amount of time to implement this properly (everything gated through a special `AlignedArray` type, which has constructors guaranteeing alignment).

On CI enabled tests using 32-byte alignment. Recent glibc versions allocate on 16-byte boundaries, so tests would never fail on my machone, even when the code was wrong. But even with 32-byte boundaries there is some non-determinism, since there is a small chance that memory that is not explicitly aligned ended up on a boundary by chance.

Advantages:

- We have explicit control over alignment.
- By using feature gating, we can use default (f32) alignment as before.
- We can give other libraries (Eigen, Tensorflow) the alignment that they need.

Disadvantages:

- As discussed, reading text files without dimensions, no requires two passes over the file.
- We have to be very careful with new code to use `Vec::with_capacity_aligned` if an embedding matrix is initialized from a `Vec`.
- Apart from non-determinism, we might miss cases where alignment is not enforced, when the code path is not covered by a unit test (I just did a grep for `vec!` and `Vec::with_capacity`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.32.0
+  - 1.36.0
   - stable
 before_script:
   - rustup component add clippy
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,12 @@ exclude = [
 
 [dependencies]
 byteorder = "1"
+cfg-if = "0.1"
 fnv = "1"
 itertools = "0.8"
 memmap = "0.7"
 ndarray = "0.12"
+num-traits = "*"
 ordered-float = "1"
 rand = "0.7"
 rand_xorshift = "0.2"
@@ -40,3 +42,7 @@ criterion = "0.3"
 name = "subword"
 harness = false
 
+[features]
+align16 = []
+align32 = []
+align64 = []
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -2,11 +2,17 @@
 
 set -ex
 
-cargo build
-cargo test
+cargo build --verbose
+cargo test --verbose
+
+# Modern glibc versions align on 16 byte boundaries, so align on
+# 32-byte boundaries to shake out errors. Unfortunately, these
+# errors can be non-deterministic (accidental allocation on a
+# boundary).
+cargo test --verbose --features "align64"
 
 # On Rust 1.32.0, we only care about passing tests.
-if rustc --version | grep -v "^rustc 1.32.0"; then
+if [ "$TRAVIS_RUST_VERSION" = "stable" ]; then
   cargo fmt --all -- --check
   cargo clippy -- -D warnings
 fi
--- a/src/align.rs
+++ b/src/align.rs
@@ -1,0 +1,260 @@
+use std::alloc::{alloc, Layout};
+use std::fmt;
+use std::iter;
+use std::marker::PhantomData;
+use std::mem;
+use std::ops::{Deref, DerefMut};
+
+use cfg_if::cfg_if;
+use ndarray::{Array, Array1, Dimension, Ix1, Ix2, ShapeBuilder, ShapeError, StrideShape};
+use num_traits::Zero;
+
+cfg_if! {
+    if #[cfg(feature = "align16")] {
+        pub type DefaultAlignment = Align16;
+    } else if #[cfg(feature = "align32")] {
+        pub type DefaultAlignment = Align32;
+    } else if #[cfg(feature = "align64")] {
+        pub type DefaultAlignment = Align64;
+    } else {
+        pub type DefaultAlignment = f32;
+    }
+}
+
+/// `AlignedArray` construction errors.
+#[derive(Debug, PartialEq)]
+pub enum AlignedArrayError {
+    /// The provided data has an incorrect alignment.
+    IncorrectAlignment,
+
+    /// ndarray shape error.
+    Shape(ShapeError),
+}
+
+impl fmt::Display for AlignedArrayError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use self::AlignedArrayError::*;
+        match *self {
+            IncorrectAlignment => write!(f, "Input array has incorrect alignment"),
+            Shape(ref err) => err.fmt(f),
+        }
+    }
+}
+
+impl std::error::Error for AlignedArrayError {
+    fn description(&self) -> &str {
+        use self::AlignedArrayError::*;
+
+        match *self {
+            IncorrectAlignment => "Input array has incorrect alignment",
+            Shape(ref err) => err.description(),
+        }
+    }
+}
+
+/// Matrix with alignment on a boundary that is a multiple of the size
+/// of type `Align`.
+pub type AlignedArray2<Align, A> = AlignedArray<Align, A, Ix2>;
+
+/// Wrapper for ndarray's `Array` that guarantees alignment on a
+/// boundary that is a multiple of the size of the type `Align`.
+pub struct AlignedArray<Align, A, D>
+where
+    D: Dimension,
+{
+    inner: Array<A, D>,
+    _phantom: PhantomData<Align>,
+}
+
+impl<Align, A> AlignedArray<Align, A, Ix1> {
+    /// Construct an `AlignedArray` from a `Vec`.
+    ///
+    /// This constructor fails when `data` is not aligned on a
+    /// boundary that is a multiple of the size of `Align`.
+    pub fn from_vec(data: Vec<A>) -> Result<Self, AlignedArrayError> {
+        if data.as_ptr().align_offset(mem::align_of::<Align>()) != 0 {
+            return Err(AlignedArrayError::IncorrectAlignment);
+        }
+
+        Ok(AlignedArray {
+            inner: Array1::from_vec(data),
+            _phantom: PhantomData,
+        })
+    }
+}
+
+impl<Align, A, D> AlignedArray<Align, A, D>
+where
+    D: Dimension,
+{
+    /// Construct an `AlignedArray`, filled with an element.
+    ///
+    /// This constructor creates an `AlignedArray` of shape `shape`,
+    /// filled with `elem`, aligned on a boundary that is a multiple
+    /// of the size of `Align`.
+    pub fn from_elem<Sh>(shape: Sh, elem: A) -> Self
+    where
+        A: Clone + Zero,
+        Sh: ShapeBuilder<Dim = D>,
+    {
+        let shape = shape.into_shape();
+
+        let mut v = Vec::with_capacity_aligned::<Align>(shape.size());
+        v.extend(iter::repeat(elem).take(shape.size()));
+
+        AlignedArray {
+            inner: Array::from_shape_vec(shape, v)
+                .map_err(AlignedArrayError::Shape)
+                .expect("Shape mismatch"),
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Construct an `AlignedArray` from a `Vec`.
+    ///
+    /// This constructor crates an `AlignedArray`, using `data` as
+    /// the backing storage with shape `shape`.
+    ///
+    /// Construction failes fails when `data` is not aligned on a
+    /// boundary that is a multiple of the size of `Align`.
+    pub fn from_shape_vec<Sh>(shape: Sh, data: Vec<A>) -> Result<Self, AlignedArrayError>
+    where
+        Sh: Into<StrideShape<D>>,
+    {
+        if data.as_ptr().align_offset(mem::align_of::<Align>()) != 0 {
+            return Err(AlignedArrayError::IncorrectAlignment);
+        }
+
+        Ok(AlignedArray {
+            inner: Array::from_shape_vec(shape, data).map_err(AlignedArrayError::Shape)?,
+            _phantom: PhantomData,
+        })
+    }
+
+    /// Construct an `AlignedArray`, filled with zeros.
+    ///
+    /// This constructor creates an `AlignedArray` of shape `shape`,
+    /// filled with zeros, aligned on a boundary that is a multiple of
+    /// the size of `Align`.
+    pub fn zeros<Sh>(shape: Sh) -> Self
+    where
+        A: Clone + Zero,
+        Sh: ShapeBuilder<Dim = D>,
+    {
+        Self::from_elem(shape, A::zero())
+    }
+}
+
+impl<Align, A, D> Deref for AlignedArray<Align, A, D>
+where
+    D: Dimension,
+{
+    type Target = Array<A, D>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl<Align, A, D> DerefMut for AlignedArray<Align, A, D>
+where
+    D: Dimension,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
+
+impl<Align, A, D> fmt::Debug for AlignedArray<Align, A, D>
+where
+    A: fmt::Debug,
+    D: Dimension,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "AlignedArrayError {{ inner: {:?}}}", self.inner)
+    }
+}
+
+/// Allocate an array type with aligned memory.
+pub trait WithCapacityAligned<T> {
+    /// Allocate an array type with aligned memory.
+    ///
+    /// The type with the length `capacity` is aligned on a boundary
+    /// that is a multiple of the size of `Align`.
+    fn with_capacity_aligned<Align>(capacity: usize) -> Self;
+}
+
+impl<T> WithCapacityAligned<T> for Vec<T> {
+    fn with_capacity_aligned<Align>(capacity: usize) -> Self {
+        // alloc() beheviour is undefined for zero-sized layouts.
+        if capacity == 0 {
+            return Vec::with_capacity(0);
+        }
+
+        let align_size = mem::size_of::<Align>();
+        let type_size = mem::size_of::<T>();
+
+        let layout = Layout::from_size_align(capacity * type_size, align_size).unwrap();
+
+        // Allocate and wrap.
+        unsafe {
+            let p = alloc(layout);
+            Vec::from_raw_parts(p as *mut T, 0, capacity)
+        }
+    }
+}
+
+#[allow(dead_code)]
+#[repr(align(16))]
+pub struct Align16 {
+    _data: [u8; 16],
+}
+
+#[cfg(feature = "align32")]
+#[repr(align(32))]
+pub struct Align32 {
+    _data: [u8; 32],
+}
+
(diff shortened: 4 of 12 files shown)


danieldk:
BTW, Tensorflow now seems to allocate on 64-byte boundaries:

https://github.com/tensorflow/tensorflow/pull/19121

danieldk:
> At least on Rust 1.38, there are compiler warnings that `Align16` is unused.

Interesting, I am not getting errors on 1.38. `Align16` also has `allow(dead_code)`.

Speaking of compiler versions, this would bump up the minimum version to 1.36.0 for `align_offset`. This method also has the worrisome remark:

*It is permissible for the implementation to always return usize::max_value(). Only your algorithm's performance can depend on getting a usable offset here, not its correctness.*

It is not clear to me whether this means with a particular alignment or not. I should probably read the RFC.

> The constructors in `align.rs` could maybe benefit from some docs, although the module is `pub (crate)`, it might make things clearer for maintenance in the future.

Yes, I marked this WIP to get some feedback on the general direction. `AlignedArray` also needs some unit tests to test rejection of unaligned `Vec`s.

> Finally a question, I'm surprised that we're not getting an error for leaking a crate-private type in a public interface through `NdArray(AlignedArray)`. From what I can tell, we're also removing the possibility to construct `NdArray` from outside the crate, since that'd require construction of the private `AlignedArray`.

I guess that we could make `AlignedArray` public. But that'd probably also mean that we need to make the `WithCapacityAligned` trait public to be useful. But then we start leaking a lot implementation details, of which we may want to change the API in the future (e.g. once ndarray support custom storages, that would probably be the better route than a wrapper type).

danieldk:
> Speaking of compiler versions, this would bump up the minimum version to 1.36.0 for `align_offset`. This method also has the worrisome remark:
> 
> _It is permissible for the implementation to always return usize::max_value(). Only your algorithm's performance can depend on getting a usable offset here, not its correctness._
> 
> It is not clear to me whether this means with a particular alignment or not. I should probably read the RFC.

The RFC is nicer:

*It is perfectly valid for an implementation to always yield `usize::max_value()` to signal that the pointer cannot be aligned.*

So, that does not affect our use case, since we want to reject memory that is unaligned (an not memory that is unaligned, but can is aligned with some offset). We are effectively only interested in the result of:

`ptr as usize % alignment == 0`

sebpuetz:
> > At least on Rust 1.38, there are compiler warnings that `Align16` is unused.
> 
> Interesting, I am not getting errors on 1.38. `Align16` also has `allow(dead_code)`.

The warning is gone now, maybe the annotation wasn't there yet?

> Speaking of compiler versions, this would bump up the minimum version to 1.36.0 for `align_offset`. This method also has the worrisome remark:
> 
> _It is permissible for the implementation to always return usize::max_value(). Only your algorithm's performance can depend on getting a usable offset here, not its correctness._
> 
> It is not clear to me whether this means with a particular alignment or not. I should probably read the RFC.

Looking at the implementation, `usize::max` seems to be returned if the pointer cannot be aligned. 

> > Finally a question, I'm surprised that we're not getting an error for leaking a crate-private type in a public interface through `NdArray(AlignedArray)`. From what I can tell, we're also removing the possibility to construct `NdArray` from outside the crate, since that'd require construction of the private `AlignedArray`.
> 
> I guess that we could make `AlignedArray` public. But that'd probably also mean that we need to make the `WithCapacityAligned` trait public to be useful. But then we start leaking a lot implementation details, of which we may want to change the API in the future (e.g. once ndarray support custom storages, that would probably be the better route than a wrapper type).

Maybe we can offer a safe API with non-finalfusion types to construct `NdArray` storages? That probably requires copying, but I guess for a rarely used constructor that'd be fine...

danieldk:
> I guess that we could make `AlignedArray` public. But that'd probably also mean that we need to make the `WithCapacityAligned` trait public to be useful. But then we start leaking a lot implementation details, of which we may want to change the API in the future (e.g. once ndarray support custom storages, that would probably be the better route than a wrapper type).

Maybe we should look if we are using the fact that `NdArray` is a tuple struct with a `pub` member somewhere outside this crate.

sebpuetz:
> > I guess that we could make `AlignedArray` public. But that'd probably also mean that we need to make the `WithCapacityAligned` trait public to be useful. But then we start leaking a lot implementation details, of which we may want to change the API in the future (e.g. once ndarray support custom storages, that would probably be the better route than a wrapper type).
> 
> Maybe we should look if we are using the fact that `NdArray` is a tuple struct with a `pub` member somewhere outside this crate.

https://github.com/finalfusion/finalfusion-python/blob/178ca4f643b8fdce1ff3a39679570337f0dc3403/src/embeddings.rs#L235 

```Rust
                StorageWrap::NdArray(array) => array.0.view(),
```
but I guess that could be delegated to a method on `NdArray`.

danieldk:
> https://github.com/finalfusion/finalfusion-python/blob/178ca4f643b8fdce1ff3a39679570337f0dc3403/src/embeddings.rs#L235
> 
> ```rust
>                 StorageWrap::NdArray(array) => array.0.view(),
> ```
> 
> but I guess that could be delegated to a method on `NdArray`.

Yes, that should just be `array.view()` or something along those lines.

danieldk:
> > https://github.com/finalfusion/finalfusion-python/blob/178ca4f643b8fdce1ff3a39679570337f0dc3403/src/embeddings.rs#L235
> > ```rust
> >                 StorageWrap::NdArray(array) => array.0.view(),
> > ```
> > 
> > 
> > but I guess that could be delegated to a method on `NdArray`.
> 
> Yes, that should just be `array.view()` or something along those lines.

Regardless of whether we merge this PR, I think `NdArray` should be changed to hide its inner type. At the time I thought it's nice so that callers can easily make their own storage by wrapping an array. But it adds a lot of inflexibility.

I'll open a PR for this.

sebpuetz:
Let me know when I should go through this one again, or are you going to open another PR with your second idea?

danieldk:
The second idea didn't pan out. I thought of writing a custom allocator. But the custom allocator API doesn't allow you to see what memory is being allocated for, so then *everything* would be aligned on 64-byte boundaries, which would be a huge waste (e.g. for the vocab strings).

However, this did make me realize that Rust actually has a publicly-accessible allocation API, which does not require `unstable`. So I rewrote the aligned `Vec` allocation stuff using this API, which simplifies things quite a bit.

I'll add some more documentation and move this PR from Draft to real. Another review then would be great. I'll do a ping.

danieldk:
@sebpuetz ok, docs added.

Before merging, it would be nice if we could actually test this with Tensorflow in some fashion. With the C API, it'd be fairly simple, `TF_NewTensor` will copy and deallocate the data when the alignment is not correct. If it is correct, it will only deallocate the data when the tensor is deallocated. Deallocation is done through a user-provided function. In our case, this should be a no-op, since Rust is responsible for cleaning up memory, but we could use this function to observe deallocation.

We cannot test this from the `tensorflow` crate, since it does not provide a `Tensor` constructor that does not clone data.

sebpuetz:
> @sebpuetz ok, docs added.

Content wise this looks fine.

> Before merging, it would be nice if we could actually test this with Tensorflow in some fashion. With the C API, it'd be fairly simple, `TF_NewTensor` will copy and deallocate the data when the alignment is not correct. If it is correct, it will only deallocate the data when the tensor is deallocated. Deallocation is done through a user-provided function. In our case, this should be a no-op, since Rust is responsible for cleaning up memory, but we could use this function to observe deallocation. But I don't know how this would/should be exposed in Tensorflow as an op (?). Another small concern that I have is the deallocation order in Tensorflow -- what if Tensorflow believes that the embedding op can be garbage collected, while it is still using the embedding matrix? (Do ops exist throughout the lifetime of the program?)
> We cannot test this from the `tensorflow` crate, since it does not provide a `Tensor` constructor that does not clone data. (It would also entail that `Tensor` needs a lifetime that is bound to the original memory.)

We'd want to expose a method in `finalfusion-ffi` which returns a pointer to the embedding matrix, then use that pointer to construct a tensor through tensorflow's C API and observe how deallocation is handled? 

Ideally, we would just use the reference counting on the embeddings resource to deallocate the embeddings matrix and leave ownership of the embedding matrix with the resource. Unfortunately I couldn't find anything in the public C++ API that would allow using the pointer from Rust directly in a tensor, let alone keeping ownership of the data. `NumpyTensorBuffer` is even declared as a friend class to get access to a private `Tensor` constructor https://github.com/tensorflow/tensorflow/blob/00fad90125b18b80fe054de1055770cfb8fe4ba3/tensorflow/core/framework/tensor.h#L639-L645. 

The `TensorBuffer` constructor taking a raw pointer is public on the other hand. We could easily implement our own `Buffer` type that wouldn't deallocate Rust memory and handle the ref counting on our resource, but I don't see a way to actually construct a tensor in C++.

There might be a workaround through using a fake allocator as suggested here https://github.com/tensorflow/tensorflow/issues/6955#issuecomment-300646846 since there is a constructor  `Tensor(Allocator* a, DataType type, const TensorShape& shape)`., `Allocator* a` is also responsible for deallocating the buffer, so we might have an angle through this approach.

danieldk:
> There might be a workaround through using a fake allocator as suggested here [tensorflow/tensorflow#6955 (comment)](https://github.com/tensorflow/tensorflow/issues/6955#issuecomment-300646846) since there is a constructor `Tensor(Allocator* a, DataType type, const TensorShape& shape)`., `Allocator* a` is also responsible for deallocating the buffer, so we might have an angle through this approach.

I think it depends on how we expose things. If we want to allow backpropagation, this should probably be an op in the graph rather than a tensor that is passed to the graph. So I guess the question is twofold: how long do ops live (as long as the session(?))? And what would this op look like? Would this be the current op, which then also allows backprop? Or would this be a separate op that just represents a matrix? In the latter case, how would users get the word -> index mapping?

danieldk:
Another issue, that this PR does not address, is alignment of the embeddings themselves. Wrapping an embedding in a tensor now also requires a copy due to alignment mismatches. This is probably a more frequent scenario than the one we are discussing now (e.g. when we'd want to pass finalfusion embeddings into through the `tensorflow` crate). However, it is more problematic for the borrowed non-subword embeddings, since they are not guaranteed to be aligned on a boundary and the only way to do so is by copying them. Solving this issue would require us to insert padding after every embedding, so that the next embedding is on a boundary again. This would not only benefit interaction between the finalfusion and tensorflow crates, but would also allow us to use aligned SIMD in the finalfusion crate (e.g. for summing subword embeddings). But it makes the goal proposed by this PR harder to archieve, since the raw embedding matrix represents embeddings + padding.

At any rate, I will probably make a `finalfusion-tensorflow` crate that provides a small wrapper around finalfusion to do lookups that result in `Tensor` objects. This would be a large boon to e.g. `finalfusion` and `dpar`.

sebpuetz:
> > There might be a workaround through using a fake allocator as suggested here [tensorflow/tensorflow#6955 (comment)](https://github.com/tensorflow/tensorflow/issues/6955#issuecomment-300646846) since there is a constructor `Tensor(Allocator* a, DataType type, const TensorShape& shape)`., `Allocator* a` is also responsible for deallocating the buffer, so we might have an angle through this approach.
> 
> I think it depends on how we expose things. I think it is good to think about the API/user experience before thinking about concrete types. E.g. if the matrix ends up being associated with the rest of the embeddings, then there is no necessity to jump through memory management hoops.

Looking at concrete types and implementation was mostly about the feasibility of using the embedding matrix directly in a TF tensor.

> We should also consider whether this is a realistic application. As we already discussed, it would be hard to optimize both the word and subword matrix, without divergence (it's possible, but we'd have to implement differentiation in the op and subtract the subword embeddings from the word embeddings, undo normalization, etc). I agree with you that it would be worthwhile to explore pretraining on a corpus, while backpropagating into embeddings as well, but things should be proven to work before we make the finalfusion implementation more complex.



> If we want to allow backpropagation, this should probably be an op in the graph rather than a tensor that is passed to the graph. So I guess the question is twofold: how long do ops live (as long as the session(?))? And what would this op look like?

When an op is created, a kernel is constructed. The constructor sets attributes of how that kernel is executed (e.g. the masking parameters in the lookup method in `finalfusion-tensorflow`). The kernels compute method sets the output of the method when computation happens. In this case I though about a similar interface to the other methods we have. The input is an initialized `Embeddings` resource and the output is a tensor being backed by the `Embeddings`' data. The lifetime of the kernel wouldn't matter in this case since all data is managed by the tensor or `Embeddings` anyways.

> Would this be the current op, which then also allows backprop? Or would this be a separate op that just represents a matrix? In the latter case, how would users get the word -> index mapping?

I might misremember this, but I thought you proposed keeping the current simple op (strings in, embeddings out) and add three new ops: one to get access to indices and one to get access to the raw matrix. Then users could use gather to get access to the desired embeddings and backprop/differentiation would happen through that avenue. Finally a write op that serializes the changed embedding matrix again.

We can find out whether divergent subword and word embeddings are an issue experimentally, but I'd propose doing so through the python interface (or perhaps the native Rust library + tensorflow crate)

> Another issue, that this PR does not address, is alignment of the embeddings themselves. Wrapping an embedding in a tensor now also requires a copy due to alignment mismatches. This is probably a more frequent scenario than the one we are discussing now (e.g. when we'd want to pass finalfusion embeddings into through the tensorflow crate). However, it is more problematic for the borrowed non-subword embeddings, since they are not guaranteed to be aligned on a boundary and the only way to do so is by copying them. Solving this issue would require us to insert padding after every embedding, so that the next embedding is on a boundary again. This would not only benefit interaction between the finalfusion and tensorflow crates, but would also allow us to use aligned SIMD in the finalfusion crate (e.g. for summing subword embeddings). But it makes the goal proposed by this PR harder to archieve, since the raw embedding matrix represents embeddings + padding.

Ah, I didn't know this, would this be an issue if we have direct access to the matrix and use gather to retrieve embeddings?

> At any rate, I will probably make a finalfusion-tensorflow crate that provides a small wrapper around finalfusion to do lookups that result in Tensor objects. This would be a large boon to e.g. finalfusion and dpar.

I'll rename the op repo later.

sebpuetz:
One more point about downstream applications like `dpar` or `sticker`, we wouldn't be handling embeddings in rust at all with the tf ops. We would only pass a string tensor to run the session, which means no copying in the tensorflow crate. Copying happens into a Rust vec, then a C++ vec and finally into the actual tensor.

danieldk:
> One more point about downstream applications like `dpar` or `sticker`, we wouldn't be handling embeddings in rust at all with the tf ops. We would only pass a string tensor to run the session, which means no copying in the tensorflow crate. Copying happens into a Rust vec, then a C++ vec and finally into the actual tensor.

I am not sure if I understand this correctly, but I wouldn't want to use the op in `dpar` or `sticker`. So, this is why I'd like to optimize the finalfusion -> tensorflow rust path.

sebpuetz:
> > One more point about downstream applications like `dpar` or `sticker`, we wouldn't be handling embeddings in rust at all with the tf ops. We would only pass a string tensor to run the session, which means no copying in the tensorflow crate. Copying happens into a Rust vec, then a C++ vec and finally into the actual tensor.
> 
> I am not sure if I understand this correctly, but I wouldn't want to use the op in `dpar` or `sticker`. So, this is why I'd like to optimize the finalfusion -> tensorflow-rust path.

I see, I thought removing the boilerplate would've been nice.

Anyways, iirc for the ffi and tensorflow op we came to the conclusion that copying would hardly be a bottleneck which is why we're returning a Vec and copy that in each layer? So why do we want to optimize this now? 

I was also under the impression that this PR was about setting up exposure of the embeddings matrix to tensorflow, so now I'm unsure about the direction we are going

danieldk:
> I see, I thought removing the boilerplate would've been nice.

But we get a lot of additional bioler plate and complexity in the build process. sticker and dpar are now pure Rust builds, which is really nice. Additionally, we'd be introducing more unsafe C code, whereas now we can do nearly everything in safe Rust.

> Anyways, iirc for the ffi and tensorflow op we came to the conclusion that copying would hardly be a bottleneck which is why we're returning a Vec and copy that in each layer? So why do we want to optimize this now?

There are two different goals here:

- If we align embeddings on boundaries, we can use unaligned SIMD instructions in finalfusion itself (e.g. when computing subword representations, doing similarity queries, etc.).

- I'd like to have an interface between finalfusion and the Tensorflow Rust crate, so that downstream users such as dpar and sticker do not need custom code to go from embeddings to Tensorflow tensors.

If we are doing that anyway, I think zero-copy would be a nice additional thing to have. This would be much easier to achieve than in the FFI. Since most things happen in Rust, we can use lifetimes to track memory. Of course, achieving full zero-copy would require changes to the `tensorflow` crate. I'll send an e-mail to the `tensorflow-rust` list in the coming days, asking whether Adam is willing to accept such changes. If so, I'll do a PR. The proposal would be to add a lifetime to the `Tensor` type, which is `'static` for newly allocated tensors, but would bind to a life time of a slice when a `Tensor` is constructed from a slice. In this way, we could safely pass Rust memory to tensorflow.
 
> I was also under the impression that this PR was about setting up exposure of the embeddings matrix to tensorflow, so now I'm unsure about the direction we are going

Yes, but then I realized that we are missing optimization opportunities by using unaligned vectors. I thought about this several months ago (when I did a quick implementation of matrix alignment), but then forgot about it.

danieldk:
I am closing this in favor of a new PR that extends this PR to also offer the possibility to align embeddings.

danieldk:
Performance comparison

```
group                           align                                  master
-----                           -----                                  ------
ndarray-lookup-allround         1.03   461.0±61.04ns        ? B/sec    1.00   447.3±56.43ns        ? B/sec
ndarray-lookup-known            1.01   234.6±29.38ns        ? B/sec    1.00   233.2±23.22ns        ? B/sec
ndarray-lookup-unknown          1.01     18.3±2.16µs        ? B/sec    1.00     18.1±2.01µs        ? B/sec
ndarray-mmap-lookup-allround    1.03  849.9±102.67ns        ? B/sec    1.00  825.6±101.60ns        ? B/sec
ndarray-mmap-lookup-known       1.01   493.3±32.29ns        ? B/sec    1.00   488.5±32.22ns        ? B/sec
ndarray-mmap-lookup-unknown     1.02     23.0±2.34µs        ? B/sec    1.00     22.6±2.26µs        ? B/sec
opq-lookup-allround             1.00    44.6±15.23µs        ? B/sec    1.00    44.6±15.44µs        ? B/sec
opq-lookup-known                1.00     20.7±0.67µs        ? B/sec    1.01     20.9±0.55µs        ? B/sec
opq-lookup-unknown              1.00  963.1±113.43µs        ? B/sec    1.01  974.7±115.92µs        ? B/sec
opq-mmap-lookup-allround        1.00    45.0±15.84µs        ? B/sec    1.00    45.0±15.42µs        ? B/sec
opq-mmap-lookup-known           1.00     20.8±0.59µs        ? B/sec    1.00     20.9±0.55µs        ? B/sec
opq-mmap-lookup-unknown         1.00  964.3±112.83µs        ? B/sec    1.01  969.7±112.98µs        ? B/sec
```

sebpuetz:
How come mmap on align is slower than on master?

PR #59, opened 2019-09-27
Change NdArray type: tuple struct -> struct

danieldk:


sebpuetz:
CI fails on stable because of some spurious newline in `similarity.rs`

PR #61, opened 2019-09-27
Add NGram Indexer.

sebpuetz:
This commit adds an NGramIndexer with explicitly stored ngrams.
Returning an index through index_ngram is now optional.

### DIFF
--- a/src/chunks/vocab.rs
+++ b/src/chunks/vocab.rs
@@ -515,19 +515,24 @@ impl Vocab for VocabWrap {
 pub trait NGramIndices {
     /// Return the subword ngrams and their indices of a word,
     /// in the subword vocabulary.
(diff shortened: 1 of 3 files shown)


danieldk:
By the way, we should probably finish up #58 before making larger changes throughout the crate, since it could lead to merge commits. This one is non-obtrusive though.

sebpuetz:
Adding full support only requires changes to `vocab.rs` and adding chunkidentifiers to `io.rs`, I believe none of that is touched by #58, but I can hold off the PRs until that's in

danieldk:
Ok, if it only touches vocab part then that's fine. By the way, I have an idea for how to tackle the alignment issue completely differently, which may be less invasive in the crate. But I have to hash out the details a bit.

PR #62, opened 2019-09-27
Add SubwordVocab with NGramIndexer.

sebpuetz:
Implement read and write methods for new vocab type.

I factored out read and write methods for words & ngrams since there'd be a lot of duplicates of that.

A new variant to `VocabWrap` would come in with the next PR, enabling full support for explicit ngram vocabs.

### DIFF
--- a/src/chunks/vocab.rs
+++ b/src/chunks/vocab.rs
@@ -10,7 +10,7 @@ use super::io::{ChunkIdentifier, ReadChunk, WriteChunk};
 use crate::compat::fasttext::FastTextIndexer;
 use crate::io::{Error, ErrorKind, Result};
 use crate::subword::{
-    BucketIndexer, FinalfusionHashIndexer, Indexer, NGramsIndices,
+    BucketIndexer, FinalfusionHashIndexer, Indexer, NGramIndexer, NGramsIndices,
     SubwordIndices as StrSubwordIndices,
 };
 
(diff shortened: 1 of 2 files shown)


src/chunks/vocab.rs
```diff
@@ -301,14 +294,90 @@ where
...
+        let min_n = read
+            .read_u32::<LittleEndian>()
+            .map_err(|e| ErrorKind::io_error("Cannot read minimum n-gram length", e))?;
+        let max_n = read
+            .read_u32::<LittleEndian>()
+            .map_err(|e| ErrorKind::io_error("Cannot read maximum n-gram length", e))?;
+
+        let words = read_vocab_items(read, words_len as usize)?;
+        let ngrams = read_vocab_items(read, ngrams_len as usize)?;
+        let indexer = NGramIndexer::new(ngrams);
```
NianhengWu: Just a question: Is this variable ```indexer``` unnecessary or it doesn't matter?

sebpuetz: Thanks for the review! 

The `indexer` is used to retrieve the ngram indices in the `SubwordVocab`. For the other vocabularies, the indexer hashes an ngram and returns the hashcode, the hashcode does not necessarily map to a unique ngram, so there could be collisions. This indexer has a `HashTable` explicitly storing ngram-index mappings so that each ngram is assigned a unique index.

NianhengWu: Thanks for the answer! Sorry I didn't make it clear. Actually I mean in cases like this, I always do something like:
```
Ok(SubwordVocab::new(
    words, 
    min_n, 
    max_n, 
    indexer: NGramIndexer::new(ngrams),
)
```
Instead of creating another variable ```indexer```, and I'm not so sure if this is a common rule or it just doesn't matter that much :)

sebpuetz: I see, the two variants are equivalent, so it's just down to readability. I think both versions are commonly used, but personally I like to only use variables in the initialization and handle construction separately.

Issue #63, opened 2019-09-29
Update README with examples


sebpuetz:
Some code snippets could be nice. Also, a synopsis of what you can do with the crate.

The README also states, that the crate is new and the API will change, while the API might still change, I think we're approaching somewhat stable levels in terms of public API.

danieldk:
> Some code snippets could be nice.

Any code snippets should go in the rustdoc, we get automatic tests, ensuring that the snippets are not outdated. Of course, the readme could contain a link to the examples section of the crate main page.

> Also, a synopsis of what you can do with the crate. The README also states, that the crate is new and the API will change, while the API might still change, I think we're approaching somewhat stable levels in terms of public API.

I agree with both points.

PR #64, opened 2019-09-29
Update README.md

sebpuetz:
Addresses part of what we discussed in #63. Adding doctests should be done separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,77 @@
 [![docs.rs](https://docs.rs/finalfusion/badge.svg)](https://docs.rs/finalfusion/)
 [![Travis CI](https://img.shields.io/travis/finalfusion/finalfusion-rust.svg)](https://travis-ci.org/finalfusion/finalfusion-rust)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,9 +4,77 @@
...
+
+`finalfusion` supports a variety of formats:
+
+* Vocabulary
+    * Subwords
+    * No subwords
+* Storage
+    * Array
+    * Memory-mapped
+    * Quantized
```
NianhengWu: Do we also need to mention memory-mapped quantized array here?

sebpuetz: I'm not sure, quantization and memory-mapping are describing different things. Memory-mapping just describes how the embedding storage can be accessed while array and quantized describe how they are stored in memory.

Maybe we can come up with a better over-all descriptions of the features? It's not really clear here how the three points can be combined, but it's what we use throughout the documentation (e.g. in `finalfusion-python` and on the webpage).

sebpuetz:
I'll merge this one for now, maybe we can think of a nicer description for storage features in a future PR!

PR #67, opened 2019-09-30
Add VocabWrap variant for NGramVocab.

sebpuetz:
Finalize support for explicit NGramVocabs.

Pretty simple stuff, I hope I didn't overlook anything.

If we could cut a release soon after this PR is merged, `finalfrontier` could rely on versions from crates.io instead of pointing to git versions temporarily.


sebpuetz:
I made `NGrams` public since we need an iterator over ngrams to construct an `NGramIndexer`.

sebpuetz:
Great, I believe that's all we need from `finalfusion-rust`. I'll go ahead and open a PR in `finalfrontier` pointing at this commit as dependency so we can wait a little longer before releasing a new `finalfusion-rust` version.

Issue #71, opened 2019-10-01
Fail CI upon benchmark failure


danieldk:
This is a trivial change, but should go in a separate PR.

sebpuetz:
I.e., move the benches to the `ci-script` instead of `after-success`?

danieldk:
I don't like that semantically. It's really not part of the main build. Also, it relies on network, which tends to fail every once in a while on Travis CI. But it would be nice to get notified somehow if this fails.

I'll look into this a bit deeper, I think this could perhaps be solved by jointly also reporting the bench results on some page or in the PR request. Because now you have to dig in the CI results to see regressions.

danieldk:
Fixed in #75.

PR #72, opened 2019-10-01
Fail CI upon benchmark failure

NianhengWu:
If I understood it correctly... #71 

Not sure about the Shebang though


NianhengWu:
> Let's use this as a stop gap. Could you rebase this on master?

I just did :) Is it working?

NianhengWu:
> You forgot to remove `after-success.sh` from `.travis.yml`.
> 
> Also, you should do this pull request from another branch than `master`, otherwise we can't see if the benchmarks run correctly (since they only run when the branch is not `mast

PR #74, opened 2019-10-02
Move bench into 'ci/script'.

NianhengWu:


danieldk:
A new interesting problem: the git hash is not valid if the PR comes from another repo? I guess that we'd have to clone the repository of the PR submitter. But that has other problems (e.g. `master` may be stable).

Another possibility would be to add the submitter's repo as an additional git upstream and fetch it, which would make the hash valid

This all seems to be a bit fragile to run as a CI-failing test :/.

danieldk:
An additional problem is that benchmarks on public CI are apparently not very reliable:

https://bheisler.github.io/post/benchmarking-in-the-cloud/

Maybe we should remove benchmarks from CI and ask people to submit a bench diff for changes that potentially affect performance? (The machine does not really matter, as long as it is idle, since we are interested in the relative change.)

NianhengWu:
https://travis-ci.org/finalfusion/finalfusion-rust/jobs/592571963#L972

> A new interesting problem: the git hash is not valid if the PR comes from another repo? I guess that we'd have to clone the repository of the PR submitter. But that has other problems (e.g. `master` may be stable).
> 
> Another possibility would be to add the submitter's repo as an additional git upstream and fetch it, which would make the hash valid
> 
> This all seems to be a bit fragile to run as a CI-failing test :/.

That's so weird... The line which failed is```git checkout ${TRAVIS_COMMIT}```,  and I checked the definition of ```TRAVIS_COMMIT```, and it says:

> TRAVIS_COMMIT: The commit that the current build is testing.

But the commit id from [here](https://travis-ci.org/finalfusion/finalfusion-rust/jobs/592571963#L971) is different from the id of the commit I just made...

NianhengWu:
> An additional problem is that benchmarks on public CI are apparently not very reliable:
> 
> https://bheisler.github.io/post/benchmarking-in-the-cloud/
> 
> Maybe we should remove benchmarks from CI and ask people to submit a bench diff for changes that potentially affect performance? (The machine does not really matter, as long as it is idle, since we are interested in the relative change.)

I like the idea :)

danieldk:
Ok, so, maybe you could make another PR (sorry ;)). Yanking the benchmarks from CI?

NianhengWu:
> Ok, so, maybe you could make another PR (sorry ;)). Yanking the benchmarks from CI?

Sure! :)

PR #76, opened 2019-10-03
Add WordIndex::word and WordIndex::subword methods

danieldk:
These will return an Option with the word index or subword indices, as
appicable.


danieldk:
> Maybe implement `IntoIterator` for `WordIndex`, too? (Not in this PR)

That would be really nice!

PR #77, opened 2019-10-06
Expose indexer and ngram range for subword vocabs.

sebpuetz:
Being able index single NGrams would be nice. In order to do so, we also need to expose a constructor for `StrWithCharLen`.

### DIFF
--- a/src/chunks/vocab.rs
+++ b/src/chunks/vocab.rs
@@ -154,6 +154,21 @@ where
         }
     }
 
+    /// Get the vocab's indexer.
+    pub fn indexer(&self) -> &I {
+        &self.indexer
+    }
+
+    /// Get the lower bound of the generated ngram lengths.
+    pub fn min_n(&self) -> u32 {
+        self.min_n
+    }
+
+    /// Get the upper bound of the generated ngram lengths.
+    pub fn max_n(&self) -> u32 {
+        self.max_n
+    }
+
     fn bracket(word: impl AsRef<str>) -> String {
         let mut bracketed = String::new();
         bracketed.push(Self::BOW);
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -172,7 +172,21 @@ pub struct StrWithCharLen<'a> {
     char_len: usize,
 }
 
+impl<'a> From<&'a str> for StrWithCharLen<'a> {
+    fn from(s: &'a str) -> Self {
+        StrWithCharLen::new(s)
+    }
+}
+
 impl<'a> StrWithCharLen<'a> {
+    /// Construct `StrWithCharLen`.
+    ///
+    /// Counts the number of chars in a `&str` and constructs a `StrWithCharLen` from it.
+    pub fn new(s: &'a str) -> Self {
+        let char_len = s.chars().count();
+        StrWithCharLen { inner: s, char_len }
+    }
+
     pub fn as_str(&self) -> &str {
         self.inner
     }


danieldk:
Can you give an example of where this would be useful? I don't really like exposing `StrWithCharLen`.

sebpuetz:
Doing ungodly stuff like:
https://github.com/sebpuetz/finalfusion-python/blob/b303ecc0268396f137d5ce0772e702ed67ca56cb/src/vocab.rs#L187-L200

I don't intend to pull in most of the changes from my fork, I'm using that for experiments. If you don't want to expose this in the public `finalfusion-rust` API, I'll just fork this repo, too.

danieldk:
It may be nice to expose the indexer, but I'll look at the current API and documentation first to see if it all makes sense.

src/subword.rs
```diff
@@ -172,7 +172,18 @@ pub struct StrWithCharLen<'a> {
     char_len: usize,
 }
 
+impl<'a> From<&'a str> for StrWithCharLen<'a> {
+    fn from(s: &'a str) -> Self {
```
danieldk: Put the implementation in `new`, call it from the `from` implementation.

src/chunks/vocab.rs
```diff
@@ -154,6 +154,16 @@ where
         }
     }
 
+    /// Get the vocab's indexer.
+    pub fn indexer(&self) -> &I {
+        &self.indexer
+    }
+
+    /// Get the lower and upper bound of the generated ngram lengths.
```
danieldk: Either the constructor should also take a tuple or we should have `min_n()` and `max_n()` methods. Or, perhaps even better now that we have recent Rust versions, `RangeInclusive`.

sebpuetz: For now two separate methods. If we want to use `RangeInclusive` we should do that in a separate PR since there's quite a few places where this would show up.

sebpuetz:
I amended the commit and removed the `StrWithCharLen`. Maybe we should first clean up the `NGramsIndices` vs. `SubwordIndices` and see what the fallout is. There might be a better concealed way to get a single ngram index - maybe we can add a method that takes `&str`, internally creates a `StrWithCharLen` and indexes it?

sebpuetz:
Sorry for spamming, but the implementation looks quite simple and straight forward, just added the method on `SubwordVocab::index_ngram`.

danieldk:
As I mentioned, I'd rather expose the indexer. The indexer can do this and it avoids having another utility method for a small use case.

I am just a bit hesitant about exposing `StrWithCharLen`. It felt a bit like a hack and wanted to look if there is other prior work in the Rust ecosystem.

sebpuetz:
> As I mentioned, I'd rather expose the indexer. The indexer can do this and it avoids having another utility method for a small use case.
>
> I am just a bit hesitant about exposing `StrWithCharLen`. It felt a bit like a hack and wanted to look if there is other prior work in the Rust ecosystem.
> 
> It's a bit annoying, because `StrWithCharLen` is necessary because of a design error (relying on `Vec`, which feeds the length to the hash first).

Not sure where to add the method specifically. I doubt we'd want to expose two `index_ngram` methods on the `Indexer` trait, where one takes `&str` and the other `StrWithCharLen`. 

I don't think the current `Indexer::index_ngram` should be public as there's no way to obtain `StrWithCharLen` from the outside anyways. It's not returned by any methods, so the method is unusable anyways. So I guess we should add a `pub (crate)` method somewhere to handle `StrWithCharLen` and change `Indexer::index_ngram` to take `&str`. Does that sound good to you?

danieldk:
> Not sure where to add the method specifically. I doubt we'd want to expose two `index_ngram` methods on the `Indexer` trait, where one takes `&str` and the other `StrWithCharLen`.
> 
> I don't think the current `Indexer::index_ngram` should be public as there's no way to obtain `StrWithCharLen` from the outside anyways.

It is already public? Otherwise it's not possible to implement the `Indexer` trait (from the outside).

> Does that sound good to you?

No, it's a public interface that other indexers can implement. The reason to take `StrWithCharLen` is that we can construct it on the fly without recounting the characters when iterating over n-grams.

As I said, I think some variant of `StrWithCharLen` should be public, but I just want to see if there is prior art in the Rust community. Just to avoid that we have a new type that does the same as what is already widely used. (Haven't found anything so far.)

sebpuetz:
> > Not sure where to add the method specifically. I doubt we'd want to expose two `index_ngram` methods on the `Indexer` trait, where one takes `&str` and the other `StrWithCharLen`.
> > I don't think the current `Indexer::index_ngram` should be public as there's no way to obtain `StrWithCharLen` from the outside anyways.
> 
> It is already public? Otherwise it's not possible to implement the `Indexer` trait (from the outside).

`Indexer` is public and so is `BucketIndexer`. This PR just adds the method to get it from the `SubwordVocab`.

> > Does that sound good to you?
> 
> No, it's a public interface that other indexers can implement. The reason to take `StrWithCharLen` is that we can construct it on the fly without recounting the characters when iterating over n-grams.
> 
> As I said, I think some variant of `StrWithCharLen` should be public, but I just want to see if there is prior art in the Rust community. Just to avoid that we have a new type that does the same as what is already widely used. (Haven't found anything so far.)

I was misreading that.

sebpuetz:
I reverted to the initial proposition with your requested changes.

PR #83, opened 2019-10-14
Combine subword traits.

### LINKED ISSUE #80: The naming of NGramIndices vs. SubwordIndices traits is confusing
Either they are both subwords or n-grams. But the current naming suggests that there is a difference.

sebpuetz:
Combine NGramsIndices and SubwordIndices trait.

Fixes #80

### DIFF
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -296,34 +296,60 @@ impl<'a> Iterator for NGrams<'a> {
     }
 }
 
-/// Extension trait for computing subword indices.
+/// Trait returning iterators over subwords and indices.
 ///
-/// Subword indexing assigns an identifier to each subword (n-gram) of a
-/// string. A subword is indexed by computing its hash and then mapping
-/// the hash to a bucket.
-///
-/// Since a non-perfect hash function is used, multiple subwords can
-/// map to the same index.
-pub trait SubwordIndices {
-    /// Return the subword indices of the subwords of a string.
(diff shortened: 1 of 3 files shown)


src/subword.rs
```diff
@@ -297,33 +297,55 @@ impl<'a> Iterator for NGrams<'a> {
 }
 
 /// Extension trait for computing subword indices.
-///
```
danieldk: This information is indeed outdated. However, it would be nice to have more than the current blurb. Maybe something about the requirements on implementors (e.g. collisions are permitted).

danieldk:
> Great!

Wanted to add: the boxed iterator probably does not have any measurable overhead, right?

sebpuetz:
This branch:

~~~
subwords-len-10-minn-3-maxn-6   time:   [474.97 ns 475.48 ns 475.97 ns]
~~~

Master:

~~~
subwords-len-10-minn-3-maxn-6    time:   [803.91 ns 805.15 ns 806.47 ns]
~~~

Other benches are within +-5%, mostly in favour of this branch. Before, we collected the indices into a vec twice. Once in `impl SubwordIncides for str` and another time in `impl SubwordIndices for SubwordVocab<I>`. Which btw is also a bit confusing, as we have two traits with the same name in different modules.

`subword::SubwordIndices` could be `SubwordIndicesIter` or something along those lines since it returns iterators.

PR #88, opened 2019-10-17
Improve descriptions of the vocabulary chunk types

danieldk:
These were not particularly descriptive (what is the difference between a *subword* and *ngram* vocabulary?).


danieldk:
Maybe `ExplicitSubwordVocab`, `BucketSubwordVocab`, and `FastTextSubwordVocab`? fastText is the foreign thing here, so I think it is justified if it does not follow the naming conventions.

PR #89, opened 2019-10-18
Refactor vocab module, rename subword vocabs.

sebpuetz:
FinalfusionSubwordVocab -> BucketSubwordVocab,
FinalfusionNGramVocab -> ExplicitSubwordVocab, NGramIndexer ->
ExplicitIndexer. Move SimpleVocab, SubwordVocab, VocabWrap to
their own modules.


sebpuetz:
I amended this commit to remove the `Clone` bound on `Vocab`. We don't seem to rely on it anywhere and it's possible to add the `Clone` bound explicitly whenever necessary.

src/chunks/vocab/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+mod subword;
+pub use subword::{
+    BucketSubwordVocab, ExplicitSubwordVocab, FastTextSubwordVocab, NGramIndices, SubwordIndices,
+    SubwordVocab,
+};
+
+mod simple;
+pub use simple::SimpleVocab;
+
+mod wrapper;
```
danieldk: We call the module `wrappers` for storage, so we should do the same here.

sebpuetz: There it's actually multiple wrappers, `StorageWrap` and `StorageViewWrap`. If you really don't like it, I can change it

sebpuetz:
Did the renaming, also caught another case of unnecessary `Clone` bounds.

src/chunks/vocab/mod.rs
danieldk: Yep, naming consistency between the modules is nice.

PR #90, opened 2019-10-18
Encapsulate NdNorms array.

sebpuetz:
Encapsulate NdNorms array, provide interface through deref to
Array1.

I also removed the `Norms` trait that I didn't even know existed before. I assume that was planned as a compatibility layer if we introduced other norm chunks. Now that we're planning to stabilize the API and don't intend to add new chunks and features I think this can go away until we'd have to break the API anyways. With only the `Norms` trait for accessing a norm, `NdNorms` is useless unless `Norms' is imported.

### DIFF
--- a/src/chunks/norms.rs
+++ b/src/chunks/norms.rs
@@ -2,6 +2,7 @@
 
 use std::io::{Read, Seek, SeekFrom, Write};
 use std::mem::size_of;
+use std::ops::Deref;
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 use ndarray::Array1;
@@ -9,12 +10,6 @@ use ndarray::Array1;
 use super::io::{ChunkIdentifier, ReadChunk, TypeId, WriteChunk};
 use crate::io::{ErrorKind, Result};
 use crate::util::padding;
-
-/// Trait for norm chunks.
-pub trait Norms {
-    /// Return the norm for the word at the given index.
-    fn norm(&self, idx: usize) -> f32;
-}
 
 /// Chunk for storing embedding l2 norms.
 ///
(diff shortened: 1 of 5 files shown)


NianhengWu:
The coding-style here looks more consistent now :)

src/chunks/norms.rs
```diff
@@ -24,11 +18,29 @@ pub trait Norms {
...
 /// normalized embedding by its orginal l2 norm.
 #[derive(Clone, Debug)]
-pub struct NdNorms(pub Array1<f32>);
+pub struct NdNorms {
+    norms: Array1<f32>,
+}
+
+impl NdNorms {
+    /// Construct new `NdNorms`.
+    pub fn new<T>(norms: T) -> Self
```
danieldk: For these simple cases, I prefer:

```rust
pub fn new<T>(norms: impl Into<Array1<f32>>) -> Self
```

Since it's a trivial constructor, maybe it's also worthwhile to implement `From`?

src/chunks/norms.rs
```diff
@@ -24,11 +18,29 @@ pub trait Norms {
...
+    where
+        T: Into<Array1<f32>>,
+    {
+        NdNorms {
+            norms: norms.into(),
+        }
+    }
+
+    /// Return the norm for the word at the given index.
```
danieldk: I am wondering whether it makes sense to implement `Deref` here (`DerefMut` is not necessary here). Then we wouldn't need any custom methods. Of course, this would require that the internal representation is always `Array1`, which we may not want to commit to?

src/chunks/norms.rs
```diff
@@ -24,11 +18,29 @@ pub trait Norms {
 /// The unnormalized embedding can be reconstructed by multiplying the
 /// normalized embedding by its orginal l2 norm.
 #[derive(Clone, Debug)]
-pub struct NdNorms(pub Array1<f32>);
+pub struct NdNorms {
+    norms: Array1<f32>,
```
danieldk: In these trivial wrapper structs, I prefer the field name `inner`.

src/chunks/norms.rs
sebpuetz: That was the intuition behind this, we might be able to provide `ArrayView1` with different backing storages. On the other hand, our constructor asks for `Into<Array1>` so we might aswell commit to `Array1`.

src/embeddings.rs
```diff
@@ -51,7 +51,7 @@ where
     pub fn new(metadata: Option<Metadata>, vocab: V, storage: S, norms: NdNorms) -> Self {
         assert_eq!(
             vocab.words_len(),
-            norms.0.len(),
+            norms.view().len(),
```
danieldk: `norms.len()`

PR #91, opened 2019-10-18
Encapsulate Metadata's data.

sebpuetz:
Restrict interface to Metadata's data, impl Deref(Mut) with
target toml::Value for Metadata.

### DIFF
--- a/src/chunks/metadata.rs
+++ b/src/chunks/metadata.rs
@@ -1,6 +1,7 @@
 //! Metadata chunks
 
 use std::io::{Read, Seek, Write};
+use std::ops::{Deref, DerefMut};
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 use toml::Value;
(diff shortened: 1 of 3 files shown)


src/chunks/metadata.rs
```diff
@@ -12,7 +13,30 @@ use crate::io::{Error, ErrorKind, ReadMetadata, Result};
 ///
 /// finalfusion metadata in TOML format.
 #[derive(Clone, Debug, PartialEq)]
-pub struct Metadata(pub Value);
+pub struct Metadata {
+    data: Value,
```
danieldk: `inner`

src/chunks/metadata.rs
```diff
@@ -12,7 +13,30 @@ use crate::io::{Error, ErrorKind, ReadMetadata, Result};
...
+}
+
+impl Metadata {
+    /// Construct new `Metadata`.
+    pub fn new(data: Value) -> Self {
```
danieldk: I think a `From` counterpart would be nice here as well.

sebpuetz: Tried to implement `TryFrom<V> where V: Serialize`, but that fails because of blanket impls...
~~~
  --> src/chunks/metadata.rs:43:1
   |
43 | impl<V> TryFrom<V> for Metadata where V: Serialize {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: conflicting implementation in crate `core`:
           - impl<T, U> std::convert::TryFrom<U> for T
             where U: std::convert::Into<T>;
~~~

sebpuetz:
Is this and #90 good to go?

PR #95, opened 2019-10-18
Add support for the implement of 'finalfusion reconstruct' subcommand

NianhengWu:
Add Reconstruct traits for QuantizeArray and for Embeddings with storage of QantizeArray

### DIFF
--- a/src/embeddings.rs
+++ b/src/embeddings.rs
@@ -15,8 +15,8 @@ use crate::chunks::io::{ChunkIdentifier, Header, MmapChunk, ReadChunk, WriteChun
 use crate::chunks::metadata::Metadata;
 use crate::chunks::norms::{NdNorms, Norms};
 use crate::chunks::storage::{
-    MmapArray, NdArray, Quantize as QuantizeStorage, QuantizedArray, Storage, StorageView,
-    StorageViewWrap, StorageWrap,
+    MmapArray, NdArray, Quantize as QuantizeStorage, QuantizedArray,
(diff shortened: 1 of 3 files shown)


src/embeddings.rs
```diff
@@ -513,6 +513,29 @@ impl<'a> Iterator for IterWithNorms<'a> {
...
+    /// This method reconstruct an embedding storage and construct a new
+    /// `Embeddings` struct with storage type of `NdArray`.
+    fn reconstruct(&self) -> Embeddings<V, NdArray>;
+}
+
+impl<V> Reconstruct<V> for Embeddings<V, QuantizedArray>
+where
+    V: Vocab + Clone,
+{
+    fn reconstruct(&self) -> Embeddings<V, NdArray> {
```
sebpuetz: I'd prefer this to take `self` instead of `&self`, that way the user of the API can decide whether they want to keep the old embeddings or drop them. I'm not sure whether this trait could be replaced by implementing `From`, we might run into problems with not having specialization. If it's possible, we could then add another method on an impl for
~~~Rust
impl<V, S> Embeddings<V, S> where V: Vocab + Clone, S: Reconstruct {
    fn reconstruct(self) -> Embeddings<V, NdArray> {...}
}

impl<V, S> From<Embeddings<V,S> for Embeddings<V, NdArray> where S: Reconstruct {
    fn from(reconstructable: Embeddings<V, S>) -> Self {
        reconstructable.reconstruct()
    }
}
~~~

PR #97, opened 2019-10-19
Ensure vocabs and explicit indexer don't contain duplicate words.

sebpuetz:
Add assertions to constructors to ensure that words are not
added twice.

### DIFF
--- a/src/chunks/vocab/subword.rs
+++ b/src/chunks/vocab/subword.rs
@@ -44,9 +44,16 @@ where
     /// Words are assigned indices in the given order. NGrams in range `(min_n..max_n)` are
     /// considered. The `indexer` is used to look up indices for the NGrams produced by this
     /// `SubwordVocab`.
+    ///
+    /// Panics when there are duplicate words.
     pub fn new(words: impl Into<Vec<String>>, min_n: u32, max_n: u32, indexer: I) -> Self {
         let words = words.into();
         let indices = create_indices(&words);
+        assert_eq!(
+            words.len(),
+            indices.len(),
+            "words contained duplicate entries."
+        );
 
         SubwordVocab {
             indices,
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -141,6 +141,9 @@ impl ExplicitIndexer {
 }
 
 impl ExplicitIndexer {
+    /// Construct a new explicit indexer.
+    ///
+    /// Panics when there are duplicate ngrams.
     pub fn new<V>(ngrams: V) -> Self
     where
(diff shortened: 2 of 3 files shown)


danieldk:
How does this interact with lossy reading? I guess it's possible that two different tokens with incorrect characters become the same with replacement.

sebpuetz:
Didn't consider that. 

In that case we should discard the additional embedding. The index is out of sync with the word list anyways and both get the same embedding. It's just a duplicated entry in the word-list without being useful in any way.

src/chunks/vocab/subword.rs
```diff
@@ -44,9 +44,16 @@ where
     /// Words are assigned indices in the given order. NGrams in range `(min_n..max_n)` are
     /// considered. The `indexer` is used to look up indices for the NGrams produced by this
     /// `SubwordVocab`.
+    ///
+    /// Panics if words are duplicated.
```
danieldk: *Panics when there are duplicate words.*

src/subword.rs
```diff
@@ -141,6 +141,9 @@ impl ExplicitIndexer {
 }
 
 impl ExplicitIndexer {
+    /// Construct a new explicit indexer.
+    ///
+    /// Panics if ngrams are duplicated.
```
danieldk: *Panics when there are duplicate n-grams.*

src/subword.rs
```diff
@@ -152,6 +155,11 @@ impl ExplicitIndexer {
             .enumerate()
             .map(|(idx, ngram)| (ngram, idx as u64))
             .collect::<HashMap<String, u64>>();
+        assert_eq!(
+            index.len(),
+            ngrams.len(),
+            "ngrams needs to contained duplicate ngrams."
```
danieldk: Seems like a typo ;).

sebpuetz:
Duplicated words currently get the index of the last occurrence, previous occurrences simply don't get indexed. In that case, `Embeddings::new` will fail one of the two assertions:

~~~Rust
        assert_eq!(
            vocab.words_len(),
            norms.0.len(),
            "Vocab and norms do not have the same length"
        );
        assert_eq!(
            vocab.vocab_len(),
            storage.shape().0,
            "Max vocab index must match number of rows in the embedding matrix."
        );
~~~

`words_len()` is calculated through the length of the `index`, `norms` should be in sync with the in-vocab partition of the storage. 

`vocab_len()` is also calculated through the length of the `index` and storage shape is going to be out of sync with this value if any words got duplicated by lossy reading.

PR #103, opened 2019-10-20
Fix a bunch of code that has byte order assumptions

danieldk:
This is split in several commits, since they changes affect different functionality.

### DIFF
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -217,8 +217,10 @@ impl<'a> Hash for StrWithCharLen<'a> {
     where
         H: Hasher,
     {
-        self.char_len.hash(hasher);
-        self.inner.chars().for_each(|ch| ch.hash(hasher));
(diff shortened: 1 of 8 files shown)


src/subword.rs
```diff
@@ -217,8 +217,10 @@ impl<'a> Hash for StrWithCharLen<'a> {
     where
         H: Hasher,
     {
-        self.char_len.hash(hasher);
-        self.inner.chars().for_each(|ch| ch.hash(hasher));
+        hasher.write(&(self.char_len as u64).to_le_bytes());
+        self.inner
+            .chars()
+            .for_each(|ch| hasher.write(&(ch as u32).to_le_bytes()));
```
sebpuetz: is `to_le_bytes()` no-op for little endian platforms?

danieldk: Yep.

sebpuetz: nice

PR #104, opened 2019-10-21
Add FreeBSD build configuration

danieldk:


danieldk:
Build and test without any extra features. OpenBLAS on FreeBSD is ooooold.

danieldk:
> Build passes, tests pass, not much else to see here, right?

Nope. Boring. Just wanted to check if you are ok with adding another platform to the tests.

What about Windows?

sebpuetz:
> What about Windows?

Travis supports rust builds for windows https://docs.travis-ci.com/user/reference/windows/#supported-languages

danieldk:
Ah, great, so we'll just use that.

Issue #87, opened 2019-10-17
Release 0.11


sebpuetz:
__edit:__ Copied the list from the other comment to get an indicator for the list ticks.

Going through the API:

 - [x] `prelude.rs`: #105
1. `prelude` re-exports `SimpleVocab` and `SubwordVocab` but not the aliases for `FinalfusionSubwordVocab` etc. I think it makes more sense to re-export the aliases and leave `SubwordVocab` in `chunks::vocab`.
2. `MMapQuantizedArray` is the only storage not re-exported.
3. `NdNorms` is the only public chunk not being re-exported.

 - [x] `chunks::storage::mod.rs`:

- [x] `chunks::storage::array.rs`: #93 
1. `NdArray` derives `Debug` but not `MmapArray`. 
2. `NdArray` should derive `Clone`

- [x] `chunks::storage::wrappers.rs`:
1. `StorageViewWrap` implies that it wraps a view while it actually wraps a _viewable_ storage. Think about renaming?

- [x] `chunks::mod.rs`:
1. Might be nice to explain the concept, right now the module docs state `finalfusion chunks`

- [x] `chunks::io.rs`: #92
Nothing public but from maintainers perspective:
1. `typeid_impl` is lacking docs, choices `1` for `u8` and `10` for `f32` seem arbitrary - iirc in order to leave room for other int and float types. Could still use some docs to make that clear once this is forgotten.

- [x] `chunks::metadata.rs`: #91
1. Encapsulate the inner `Value`.
2. Alternatives to lock in `toml` as our choice? Upside of `toml` is that we get easy serialization and heterogeneous collections. Downside is that we always need `Value`s to construct.

- [x] `chunks::norms.rs`: #90
1. Encapsulate `NdNorms`' inner array
2. ~~`impl Index for NdNorms` since `Norms` seems to be just that but without the `[ ]`-indexing~~ `Index::index` returns references. I'd still like to make `norm` a method directly on `NdNorms` since `NdNorms` is entirely useless without importing `Norms` otherwise.

- [x] `chunks::vocab.rs`: #89 
1. Rename vocabtypes as discussed
2. Think about reorganizing in separate modules, i.e. `vocab::mod.rs`, `vocab::subword.rs`, `vocab::simple.rs`. 
3. [x] Remove `Clone` from `Vocab`'s requirements and the other places where it pops up because of this requirement. (e.g. `Indexer` bounds)
 
- [x] `compat::fasttext`
[...]

PR #105, opened 2019-10-21
Add exports to prelude.

sebpuetz:
Point 1 in #87


src/prelude.rs
```diff
@@ -3,10 +3,16 @@
...
+    MmapArray, MmapQuantizedArray, NdArray, QuantizedArray, Storage, StorageView, StorageViewWrap,
+    StorageWrap,
 };
 
-pub use crate::chunks::vocab::{SimpleVocab, SubwordVocab, Vocab, VocabWrap};
+pub use crate::chunks::norms::NdNorms;
+
+pub use crate::chunks::vocab::{
+    BucketSubwordVocab, ExplicitSubwordVocab, FastTextSubwordVocab, SimpleVocab, Vocab, VocabWrap,
+    WordIndex,
```
danieldk: Is `WordIndex` needed?

danieldk: But a `prelude` should not export every possible symbol. Normal usage is to load an embedding matrix and to look up an embedding. The symbols for doing that should be in the `prelude`. In that sense, the `prelude` is already bloated. It's find if people have to import specific modules for more fine-grained functionality.

(And yes, all the specific vocabs/storages should probably not be in the prelude either, only: the traits and the wrappers (since people will usually want to read as `Embeddings<VocabWrap, StorageWrap>` or `Embeddings<VocabWrap, StorageViewWrap>`).

sebpuetz: I usually end up importing from between 3 to 6 modules - in addition to prelude - when using `finalfusion` as a dependency, the motivation here was to cut down on that. Maybe we can flatten the structure and re-export symbols in `chunks` from their respective submodules.

PR #107, opened 2019-10-21
Make the prelude great again

danieldk:
Remove specialized functionality that are not required for basic crate
use.

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -1,24 +1,81 @@
 //! Prelude exports the most commonly-used types and traits.
 
-pub use crate::chunks::metadata::Metadata;
+pub use crate::chunks::storage::{StorageViewWrap, StorageWrap};
 
-pub use crate::chunks::storage::{
-    MmapArray, MmapQuantizedArray, NdArray, QuantizedArray, Storage, StorageView, StorageViewWrap,
-    StorageWrap,
-};
-
-pub use crate::chunks::norms::NdNorms;
-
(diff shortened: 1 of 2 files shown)


src/prelude.rs
```diff
@@ -1,24 +1,80 @@
...
-pub use crate::embeddings::{Embeddings, Quantize};
+    #[test]
+    fn prelude_allows_reading_text_dims() {
+        let mut reader = BufReader::new(File::open("testdata/similarity.txt").unwrap());
+        Embeddings::read_text_dims(&mut reader).unwrap();
+    }
 
-pub use crate::io::{MmapEmbeddings, ReadEmbeddings, ReadMetadata, WriteEmbeddings};
+    #[test]
+    fn prelude_allows_reading_wor2vec() {
```
sebpuetz: `word2vec`

danieldk:
> What about flattening the hierarchy of the exported symbols? E.g. make `chunks` disappear in the API since that's an implementation detail.

So, you mean that we have

```
finalfusion::storage
finalfusion::vocab
```

etc. in place of:

```
finalfusion::chunks::storage
finalfusion::chunks::vocab
```

?

Completely in favor.

sebpuetz:
Yup, will do the PR once this is through.

sebpuetz:
Looks like there's a missing gate for a test?

danieldk:
> Looks like there's a missing gate for a test?

Yup, no `StorageView` for mmap on big endian.

PR #108, opened 2019-10-21
Make the API flat again.

sebpuetz:
Away with the `chunks`.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -120,7 +120,7 @@ where
 /// use std::fs::File;
 ///
 /// use finalfusion::prelude::*;
-/// use finalfusion::chunks::metadata::Metadata;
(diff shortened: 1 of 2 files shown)


danieldk:
Oh, I think I have found a workaround.

danieldk:
Doesn't work. But this seems like a bug in rustdoc, if the full module path is not visible, a symbol is made native to the module rather than a re-export. E.g. in:

```rust
pub use crate::compat::fasttext::ReadFastText;

pub use crate::embeddings::Embeddings;
```

`Embeddings` is correctly shown as re-exported. `ReadFastText` is shown as local to the prelude, because its full module path (`compat::fasttext::io::ReadFastText`) is not public. Which is a bit silly, because we expose it and import it as `compat::fasttext::ReadFastText`.

PR #116, opened 2019-10-30
Build CI pipeline using GitHub Actions

danieldk:
This change replaces Travis CI by GitHub Actions. Benefits:

- Better integrated with GitHub.
- GitHub Actions allows more 3rd party extensions (see the CI configuration for an example, no scripting needed, thanks to the 3rd party `actions-rs` extensions).
- Builds seem to be a bit faster?


danieldk:
Future PR: we should also run tests with OpenBLAS. It seems that we were not doing that with Travis CI either? (The packages were installed, but we never use the feature flags.)

danieldk:
> Support for rust looks better, too?

It does. Cross-building, etc. is much simpler.

> Do they also support builds with multiple languages?

Shouldn't be a problem - you just add the actions that set up the necessary environment. You can also run custom commands when necessary.

PR #117, opened 2019-10-30
Keep benches updated

NianhengWu:
Although we disabled benchmarks in CI, the benches are still here, so I figure this need to be fixed.


NianhengWu:
> I don't think we want the Cargo.lock in here, I think that came in with the last push?

Sorry my hand slipped and added the wrong file

PR #118, opened 2019-11-01
Implement size_hint for ngrams and optimize unknown lookup.

sebpuetz:
For BucketIndexer, the number of ngrams (and thus subword indices)
can be precomputed. Allocating a correctly sized buffer for
the subwords yields large boosts in benches.

____________________________

This led to more than 60% improvement across unknown lookups for all storages on my machine. For some reason, known-lookup also improved by ~6% across the board.

Subword benches were down ~10%.

______________

I would've preferred to have seperate `impl<I> SubwordIndices for SubwordVocab<I> where I: BucketIndexer` and `SubwordVocab<ExplicitIndexer>` but that doesn't work because of missing specialization.

~~~Rust
let size = indices.size_hint().1.unwrap();
~~~
This is a bit of a hack since `SubwordIndices` on `str` wraps the `NGramsIndicesIter` in `FilterMap` which zeroes out the lower bound because of possible filtering. But it leaves the upper bounds untouched and we know that it's the correct size for `BucketIndexer`s since those never fail to produce.

### DIFF
--- a/src/chunks/vocab/subword.rs
+++ b/src/chunks/vocab/subword.rs
@@ -13,6 +13,7 @@ use crate::subword::{
     BucketIndexer, ExplicitIndexer, FinalfusionHashIndexer, Indexer,
     SubwordIndices as StrSubwordIndices,
 };
+use crate::util::CollectWithCapacity;
 
 /// fastText vocabulary with hashed n-grams.
 pub type FastTextSubwordVocab = SubwordVocab<FastTextIndexer>;
@@ -81,7 +82,7 @@ where
     }
 
     fn bracket(word: impl AsRef<str>) -> String {
-        let mut bracketed = String::new();
+        let mut bracketed = String::with_capacity(word.as_ref().len() + 2);
         bracketed.push(Self::BOW);
         bracketed.push_str(word.as_ref());
         bracketed.push(Self::EOW);
@@ -164,11 +165,16 @@ where
     I: Indexer,
 {
     fn subword_indices(&self, word: &str) -> Option<Vec<usize>> {
-        let indices = Self::bracket(word)
+        let word = Self::bracket(word);
+        let indices = word
             .as_str()
             .subword_indices(self.min_n as usize, self.max_n as usize, &self.indexer)
-            .map(|idx| idx as usize + self.words_len())
-            .collect::<Vec<_>>();
+            .map(|idx| idx as usize + self.words_len());
+        if I::infallible() {
+            let size = indices.size_hint().1.unwrap();
+            return Some(indices.collect_with_capacity(size));
+        }
+        let indices = indices.collect::<Vec<_>>();
         if indices.is_empty() {
             None
         } else {
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -21,6 +21,9 @@ pub trait Indexer {
 
     /// Return the (exclusive) upper bound of this indexer.
     fn upper_bound(&self) -> u64;
+
+    /// Indicates whether this Indexer never fails to produce an index.
+    fn infallible() -> bool;
 }
 
 /// N-Gram indexer with bucketing.
@@ -116,6 +119,10 @@ where
         // max val is <= 64
         2u64.pow(self.buckets_exp as u32)
     }
+
+    fn infallible() -> bool {
+        true
+    }
 }
 
 impl<H> PartialEq for HashIndexer<H> {
@@ -208,6 +215,10 @@ impl Indexer for ExplicitIndexer {
 
     fn upper_bound(&self) -> u64 {
         self.bound as u64
+    }
+
+    fn infallible() -> bool {
+        false
     }
 }
 
@@ -341,6 +352,12 @@ impl<'a> Iterator for NGrams<'a> {
 
         Some(ngram_with_len)
     }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let cap_approx = (self.max_n - self.min_n + 1) * self.char_offsets.len();
+        (cap_approx, Some(cap_approx))
+    }
 }
 
 /// Trait returning iterators over subwords and indices.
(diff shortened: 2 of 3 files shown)


src/subword.rs
```diff
@@ -341,6 +352,19 @@ impl<'a> Iterator for NGrams<'a> {
 
         Some(ngram_with_len)
     }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let stop = cmp::min(self.char_offsets.len(), self.max_n);
+        let l = (self.min_n..=stop)
```
danieldk: Seems like there is a closed form?

https://www.wolframalpha.com/input/?i=sum+n-%28j-1%29%2C+j%3Da+to+b

src/chunks/vocab/subword.rs
```diff
@@ -164,11 +165,16 @@ where
...
     fn subword_indices(&self, word: &str) -> Option<Vec<usize>> {
-        let indices = Self::bracket(word)
+        let word = Self::bracket(word);
+        let indices = word
             .as_str()
             .subword_indices(self.min_n as usize, self.max_n as usize, &self.indexer)
-            .map(|idx| idx as usize + self.words_len())
-            .collect::<Vec<_>>();
+            .map(|idx| idx as usize + self.words_len());
+        if I::infallible() {
```
danieldk: Did you try how well it works to always collect with capacity? Since we are able to properly give the lower and upper bounds?

sebpuetz: The upper bounds can be off quite a lot with the explicit indexer. We also currently don't bench explicit ngrams, so my guess is we won't observe any difference in the benches.

The lower bound is guaranteed to be 0 because of `filter_map`, so that's unfortunately of no use for us.

sebpuetz:
`cargo bench` results:
~~~
ndarray-lookup-allround time:   [403.95 ns 407.19 ns 410.49 ns]                                    
                        change: [-2.7269% +0.5924% +4.0129%] (p = 0.73 > 0.05)
                        No change in performance detected.

ndarray-lookup-known    time:   [207.91 ns 209.26 ns 210.65 ns]                                 
                        change: [-6.4958% -4.2295% -1.7030%] (p = 0.00 < 0.05)
                        Performance has improved.

ndarray-lookup-unknown  time:   [6.3858 us 6.4275 us 6.4696 us]                                    
                        change: [-63.019% -61.678% -59.881%] (p = 0.00 < 0.05)
                        Performance has improved.

ndarray-mmap-lookup-allround                                                                             
                        time:   [710.31 ns 719.02 ns 727.42 ns]
                        change: [-3.0361% +0.0537% +3.3253%] (p = 0.97 > 0.05)
                        No change in performance detected.

ndarray-mmap-lookup-known                                                                            
                        time:   [419.85 ns 422.57 ns 425.32 ns]
                        change: [-2.9416% -1.1148% +0.6993%] (p = 0.25 > 0.05)
                        No change in performance detected.

ndarray-mmap-lookup-unknown                                                                             
                        time:   [7.7926 us 7.8753 us 7.9818 us]
                        change: [-62.364% -61.089% -59.583%] (p = 0.00 < 0.05)
                        Performance has improved.

opq-lookup-allround     time:   [24.079 us 25.728 us 27.522 us]                                 
                        change: [-8.2583% +1.3752% +12.523%] (p = 0.79 > 0.05)
                        No change in performance detected.

opq-lookup-known        time:   [12.058 us 12.097 us 12.147 us]                              
                        change: [-1.8102% -1.1593% -0.3846%] (p = 0.00 < 0.05)
                        Change within noise threshold.

opq-lookup-unknown      time:   [203.16 us 217.21 us 231.81 us]                               
                        change: [-65.148% -62.967% -60.638%] (p = 0.00 < 0.05)
                        Performance has improved.

opq-mmap-lookup-allround                                                                             
                        time:   [23.825 us 25.452 us 27.253 us]
                        change: [-7.8104% +1.5037% +12.586%] (p = 0.77 > 0.05)
                        No change in performance detected.

opq-mmap-lookup-known   time:   [12.014 us 12.048 us 12.093 us]                                   
                        change: [-2.6968% -1.9076% -1.1167%] (p = 0.00 < 0.05)
                        Performance has improved.

opq-mmap-lookup-unknown time:   [202.61 us 216.59 us 231.10 us]                                    
                        change: [-65.693% -63.470% -61.257%] (p = 0.00 < 0.05)
                        Performance has improved.

subwords-len-10-minn-3-maxn-6 
                        time:   [431.85 ns 432.52 ns 433.14 ns]
                        change: [-10.084% -9.8709% -9.6475%] (p = 0.00 < 0.05)
                        Performance has improved.
~~~

src/subword.rs
danieldk: I also have a vague recollection that we already used a closed form for this somewhere else.

sebpuetz: I remember discussing this with Marko while working on the slicing assignment for the Rust course, so mightve been in that context?

danieldk: Could be! In my example solution I overallocated a bit

```
let cap_approx = (n_max - n_min + 1) * seq.len();
```

Maybe I had the exact computation first, but replaced it because it detracted too much from the rest of the solution.

sebpuetz: I think there are some complications if `min_n > seq.len()` with the closed form from above. Not sure if I misread something but I think this should be it:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7c4ec5c493b63c85ee68787c2e1ead93

This returns 1, although from a sequence of length 2, we wouldn't get any ngrams. Might also be less desirable because there's a bunch of floating point operations rather than simple integer operations. Do you prefer the sum or the approximation? Or should we go for the exact calculation?

sebpuetz:
~~~Rust
        let n = self.char_offsets.len();
        if self.min_n  > n {
            (0, Some(0))
        } else {
            let a = self.min_n as f32;
            let b = cmp::min(n, self.max_n) as f32;
            let l = (0.5*(-1. + a - b)*(a + b - 2. * (1. + n as f32))) as usize;
            (l, Some(l))
        }
~~~

This seems to work but is 10% slower on the subwords bench compared to the sum version. Your approximation is on par with summing.

I'd be in favor with sticking to my initially proposed version since that is probably the easiest to read.

danieldk: If the opportunistic approximation is as fast, I would prefer that. It is constant time and provides better inlining opportunities. These are small allocations anyway, where some over-allocation does not hurt (in contrast to amortized constant appends).

PR #119, opened 2019-11-01
Share strings in vocabs. 

sebpuetz:
This is not intended to be a proper PR but rather a quick idea how we could reduce the memory footprint.

Our vocabularies are immutable, so it should be save to use references to the buffers of the words in the  `Vec<String>` in the index. If I'm not missing something, this would be a cheap way to more or less half the memory requirements for our vocabularies - the same could be done for explicit ngram vocabularies, I'd expect the impact to be bigger there.

I didn't check for exact differences, I just wanted to get some feedback whether we even want to dabble in unsafe territory and whether what I wrote is sound.

### DIFF
--- a/src/chunks/vocab/mod.rs
+++ b/src/chunks/vocab/mod.rs
@@ -2,6 +2,8 @@
 
 use std::collections::HashMap;
 use std::io::{Read, Seek, Write};
+use std::slice;
+use std::str;
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 
@@ -65,13 +67,15 @@ impl WordIndex {
     }
 }
 
-pub(crate) fn create_indices(words: &[String]) -> HashMap<String, usize> {
-    let mut indices = HashMap::new();
-
+pub(crate) fn create_indices<'a, 'b>(words: &'b [String]) -> HashMap<&'a str, usize> {
+    let mut indices = HashMap::with_capacity(words.len());
     for (idx, word) in words.iter().enumerate() {
-        indices.insert(word.to_owned(), idx);
+        unsafe {
+            let bytes = slice::from_raw_parts(word.as_ptr(), word.as_bytes().len());
+            let word = str::from_utf8_unchecked(bytes);
+            indices.insert(word, idx);
+        }
     }
-
     indices
 }
 
--- a/src/chunks/vocab/simple.rs
+++ b/src/chunks/vocab/simple.rs
@@ -11,7 +11,7 @@ use crate::io::{ErrorKind, Result};
 /// Vocabulary without subword units.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct SimpleVocab {
-    indices: HashMap<String, usize>,
+    indices: HashMap<&'static str, usize>,
     words: Vec<String>,
 }
 
@@ -24,6 +24,7 @@ impl SimpleVocab {
     pub fn new(words: impl Into<Vec<String>>) -> Self {
         let words = words.into();
         let indices = create_indices(&words);
+
         assert_eq!(
             words.len(),
             indices.len(),
--- a/src/chunks/vocab/subword.rs
+++ b/src/chunks/vocab/subword.rs
@@ -28,7 +28,7 @@ pub type ExplicitSubwordVocab = SubwordVocab<ExplicitIndexer>;
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct SubwordVocab<I> {
     indexer: I,
-    indices: HashMap<String, usize>,
+    indices: HashMap<&'static str, usize>,
     words: Vec<String>,
     min_n: u32,
     max_n: u32,
--- a/src/chunks/vocab/subword.rs
+++ b/src/chunks/vocab/subword.rs
@@ -13,6 +13,7 @@ use crate::subword::{
     BucketIndexer, ExplicitIndexer, FinalfusionHashIndexer, Indexer,
     SubwordIndices as StrSubwordIndices,
 };
+use crate::util::CollectWithCapacity;
 
 /// fastText vocabulary with hashed n-grams.
 pub type FastTextSubwordVocab = SubwordVocab<FastTextIndexer>;
@@ -81,7 +82,7 @@ where
     }
 
     fn bracket(word: impl AsRef<str>) -> String {
-        let mut bracketed = String::new();
+        let mut bracketed = String::with_capacity(word.as_ref().len() + 2);
         bracketed.push(Self::BOW);
         bracketed.push_str(word.as_ref());
         bracketed.push(Self::EOW);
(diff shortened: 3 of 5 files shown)


src/chunks/vocab/mod.rs
```diff
@@ -65,13 +67,15 @@ impl WordIndex {
     }
 }
 
-pub(crate) fn create_indices(words: &[String]) -> HashMap<String, usize> {
-    let mut indices = HashMap::new();
-
+pub(crate) fn create_indices<'a, 'b>(words: &'b [String]) -> HashMap<&'a str, usize> {
```
danieldk: In order for this to be safe, the lifetime of the strings in the `HashMap` should be tied to that of the strings in `words`, otherwise, it would be possible to deallocate the strings in `words` and the pointers in the hashmap would be wild pointers. In this implementation, it is definitely unsafe, because the strings in the `HashMap` have a static lifetime.

sebpuetz: I know that there's a mismatch but since `index` is only used internally by the vocabularies and cannot be accessed in any way from the outside, the fake-static lifetime shouldn't cause any issues, right? Or does this cause the pointers not to be dropped when the vocab (and index) go out of scope?

Do you see another way or do you know of a crate that does this properly?

danieldk: > I know that there's a mismatch but since `index` is only used internally by the vocabularies and cannot be accessed in any way from the outside, the fake-static lifetime shouldn't cause any issues, right?

It does cause issues. Clone the vocabulary, let the original go out of scope, your cloned vocab has dead pointers (to the original `String`s. So, you have to implement Clone yourself. And that's the problem, once you remove this safety, with every API change you have to take into account whether there is a situation where it could result in wild pointers. These are exactly the kinds of issues that the Rust ownership system prevents.

If we want the memory savings (for regular vocabularies it's not so pressing), use proper sharing (`Rc`, but we cannot do that since it is not `Send + Sync`, and `Arc` is probably to heavy-handed) or a proper bijective datastructure, such as a perfect hash automaton. The poor man's alternative would be a lexicographically sorted list of words, though that would require a binary search, whereas a perfect hash automaton is linear in the number of bytes/characters).

Of course, you could also implement something like `LinkedHashMap` in Rust with `unsafe`, which would be a better solution, since it does not require two members where one is dependent on the memory of the other. There is in fact such a crate, but it is largely abandoned.

At any rate, before making any savings, I am not convinced that it is really worth it. Word embeddings are typically 300 dimensions * 4 bytes = 1200 bytes. The average heap-allocated token/n-gram is almost two orders of magnitude smaller. It would only make much sense for quantized embeddings, where each embedding (with 300 dimensional embeddings) would be 150 bytes. But then you'd probably save quite a bit more through prefix/suffix sharing of dictionary automata than this type of string sharing.

sebpuetz: > It does cause issues. Clone the vocabulary, let the original go out of scope, your cloned vocab has dead pointers (to the original Strings. So, you have to implement Clone yourself. And that's the problem, once you remove this safety, with every API change you have to take into account whether there is a situation where it could result in wild pointers. These are exactly the kinds of issues that the Rust ownership system prevents.

Thanks for the explanation, I didn't take cloning into account. That is the reason why I made this draft PR, to make sure that I wasn't missing any scenarios.

> If we want the memory savings (for regular vocabularies it's not so pressing), use proper sharing (Rc, but we cannot do that since it is not Send + Sync, and Arc is probably to heavy-handed) or a proper bijective datastructure, such as a perfect hash automaton. The poor man's alternative would be a lexicographically sorted list of words, though that would require a binary search, whereas a perfect hash automaton is linear in the number of bytes/characters).

I was looking into hash-automata before but abandoned it for some reason that I forgot by now. Iirc, https://docs.rs/fst/0.3.5/fst/ required insertion in lexicographic order, so I guess the index wouldn't be frequency sorted anymore.

> Of course, you could also implement something like LinkedHashMap in Rust with unsafe, which would be a better solution, since it does not require two members where one is dependent on the memory of the other. There is in fact such a crate, but it is largely abandoned.

https://github.com/bluss/indexmap looks maintained and seems to offer a list-map hybrid. But I think any of these solutions mean a change to the API since we can't offer a view into the list as `&[String]` since the values probably aren't placed in a dense vector.

> At any rate, before making any savings, I am not convinced that it is really worth it. Word embeddings are typically 300 dimensions * 4 bytes = 1200 bytes. The average heap-allocated token/n-gram is almost two orders of magnitude smaller. It would only make much sense for quantized embeddings, where each embedding (with 300 dimensional embeddings) would be 150 bytes. But then you'd probably save quite a bit more through prefix/suffix sharing of dictionary automata than this type of string sharing.

I know that the gains would be rather modest, but it looked like a nice opportunity to cut a bit of the memory requirements.

danieldk: > I was looking into hash-automata before but abandoned it for some reason that I forgot by now. Iirc, https://docs.rs/fst/0.3.5/fst/ required insertion in lexicographic order, so I guess the index wouldn't be frequency sorted anymore.

IIRC `fst` only gives you a string -> int mapping and not an int -> string mapping, because it is not a perfect hash automaton (it stores the actual value). In contrast to e.g. my Java [Dictomaton](https://github.com/danieldk/dictomaton) library, which stores right-language cardinalities in states or transitions, which allows you to go from numbers to strings.

A downside would indeed be that strings are sorted in lexicographical order, so you either need to reshuffle the rows (expensive) or add some indirections (which adds more memory again, probably not worth it). This was one of the reasons my initial plans was to also support some on-disk perfect hash automaton format. But I put that on hold because it puts a lot of burden on someone that wants to implement finalfusion in another language.

> https://github.com/bluss/indexmap looks maintained and seems to offer a list-map hybrid. But I think any of these solutions mean a change to the API since we can't offer a view into the list as `&[String]` since the values probably aren't placed in a dense vector.

If it's made by bluss, it's acceptable ;). But indeed, we can't do all of these without API changes.

I see why you want to tackle low-hanging fruit. But I think more can be gained by reducing the size of the embedding matrices. Besides quantization and pruning, I guess it would be interesting to look how half-precision matrices perform, since they'd save half of the memory in non-quantized embeddings. Of course, in the API we'd still need to expose `f32`.

sebpuetz:
I'll close this for now since the options seem to be:
- complicate the API through manual clone
- break the API by removing slices

In retrospect, some ordered map could've been a good choice for the vocabularies since we don't actually need the slices, linear iteration probably would've been fine. But the price we're paying is not that high, so I guess it's fine. 

For explicit lookups with mincount 30, we get something like 2.8m*2 strings, but then again, if RAM is an issue, the couple 100MB from the strings probably won't matter in the light of the 2.8m embeddings.

> A downside would indeed be that strings are sorted in lexicographical order, so you either need to reshuffle the rows (expensive) or add some indirections (which adds more memory again, probably not worth it). This was one of the reasons my initial plans was to also support some on-disk perfect hash automaton format. But I put that on hold because it puts a lot of burden on someone that wants to implement finalfusion in another language.

Little off-topic note here, the format is very straight forward to implement in other languages. I implemented a native Python version on Friday afternoon, didn't implement quantized storages yet, but the rest was just a matter of an afternoon.

PR #124, opened 2019-11-16
Fix benchmarks

danieldk:


danieldk:
I guess we should look if we can just build them without running them.

sebpuetz:
We don't intend to change the API anymore anyways, so I think we probably won't have to fix them again anyways

danieldk:
Interesting, this is merged, but GitHub shows for `master` that Nicole's commit from 5 days ago is the last commit.

sebpuetz:
My activity feed also shows that you pushed Nicole's commit 5 hours ago:
image

danieldk:
Idd, because with branch protection we should not be able to go backwards.

PR #127, opened 2019-12-01
Make SubwordVocab internal fields pub for direct creation

jguhlin:
Maybe not the best way, definitely open to other suggestions, but this is the route I'm using to create SubwordVocab structs directly. Would it be a problem to merge this in?


src/chunks/vocab/subword.rs
```diff
@@ -27,11 +27,11 @@ pub type ExplicitSubwordVocab = SubwordVocab<ExplicitIndexer>;
 /// Vocabulary with subword units.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct SubwordVocab<I> {
-    indexer: I,
-    indices: HashMap<String, usize>,
-    words: Vec<String>,
-    min_n: u32,
-    max_n: u32,
+    pub indexer: I,
```
danieldk: The struct fields are implementation details and should be kept hidden, so that we can change the implementation in the future without changing the API.

It would be best to offer a constructor that takes: an indexer, the words, min_n, and max_n (`indices` can be constructed from `words`).

Cargo.lock
```diff
@@ -0,0 +1,717 @@
+# This file is automatically @generated by Cargo.
```
danieldk: `Cargo.lock` should not be added to library crates.

src/chunks/vocab/subword.rs
jguhlin: Thanks, I'll get to work on this and submit another PR (hopefully by next week).

Cargo.lock
jguhlin: Oops, embarrassing!

danieldk:
I'll close this for now. Feel free to reopen!

PR #131, opened 2020-04-15
Modernize and improve error handling

danieldk:
- Merge the `Error` and `ErrorKind` enums.
- Move the `Error` enum to the `error` module.
- Derive trait implementations using the `thiserror` crate.
- Make the `Error` enum non-exhaustive
- Replace the `ChunkIdentifier::try_from` method by an implementation
  of the `TryFrom` crate.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         rust:
           - stable
-          - 1.37.0
+          - 1.40.0
     steps:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
@@ -49,7 +49,7 @@ jobs:
       matrix:
         rust:
           - stable
-          - 1.37.0
+          - 1.40.0
     steps:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ rand = "0.7"
 rand_xorshift = "0.2"
 reductive = "0.4"
 serde = { version = "1", features = ["derive"] }
+thiserror = "1"
 toml = "0.5"
 
 [dependencies.memmap]
--- a/src/chunks/io.rs
+++ b/src/chunks/io.rs
@@ -1,10 +1,11 @@
+use std::convert::TryFrom;
 use std::fmt::{self, Display};
 use std::fs::File;
 use std::io::{BufReader, Read, Seek, Write};
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 
-use crate::io::{Error, ErrorKind, Result};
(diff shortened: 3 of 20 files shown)


sebpuetz:
>     * Make the `Error` enum non-exhaustive

That came with 1.40, so we'd have to increment the min version. If we eventually release a new version, we'd have to do the same for utils.

danieldk:
> That came with 1.40, so we'd have to increment the min version. If we eventually release a new version, we'd have to do the same for utils.

Yep. We should probably do a new version once this is merged.

danieldk:
> Although something to consider: Putting `non_exhaustive` on `Error` prevents others from re-implementing even part of the finalfusion interface based in the traits e.g. `ReadChunk`, `WriteChunk`, `MmapChunk` that we define since all of those return `Error` which is now un-implementable from the outside.

I am not sure I understand why this change changes external implementability. You can still use the `Error` enum variants before, you just can't do exhaustive pattern matching against errors? (Outside using a wildcard pattern.) Which is probably a good thing, because it allows you to add new error variants without breaking the API.

BTW, the chunk traits (`ReadChunk`, `WriteChunk` and `MmapChunk`) are currently crate-private, so it's not possible implement them.

> Fwiw, this is an API breaking change since we're changing the signature of our `Result`, so callers would be expecting `Result<T, finalfusion::io::Error>` which is different from the new `Error` type?

Yes, it's an API change, which is unfortunate :(. But I think it's a relatively minor hassle that buys more API stability in the future (through `non_exhaustive`).

sebpuetz:
> > Although something to consider: Putting `non_exhaustive` on `Error` prevents others from re-implementing even part of the finalfusion interface based in the traits e.g. `ReadChunk`, `WriteChunk`, `MmapChunk` that we define since all of those return `Error` which is now un-implementable from the outside.
> 
> I am not sure I understand why this change changes external implementability. You can still use the `Error` enum variants like before, you just can't do exhaustive pattern matching against `Error`? (Outside using a wildcard pattern.) Which is a good thing, because it allows us to add new error variants without breaking the API.
>
> BTW, the chunk traits (`ReadChunk`, `WriteChunk` and `MmapChunk`) are currently crate-private, so it's not possible implement them.

Guess my reading comprehension failed: Only structs and enum **variants** can't be created from external crates if annotated with `non_exhaustive`. Doesn't apply to `non_exhaustive` enums, just `non_exhaustive` **variants**. So, my considerations were pretty pointless ;).

> > Fwiw, this is an API breaking change since we're changing the signature of our `Result`, so callers would be expecting `Result<T, finalfusion::io::Error>` which is different from the new `Error` type?
> 
> Yes, it's an API change, which is unfortunate :(. But I think it's a relatively minor hassle that buys more API stability in the future (through `non_exhaustive`).

Sure, just wanted to point this out since we put a claim in the README that we're API stable.

PR #132, opened 2020-04-15
Bump finalfusion to 0.12.0 to mark the error API change

danieldk:


sebpuetz:
Moving mmap stuff to a feature gated `mmap` mod shouldn't be API changing, right? If not, then this should be ready to go and creating a `mmap` mod could be done in a point release.

Otherwise I could put some time (tonight/tomorrow) into reshuffling the modules before this release.

danieldk:
> Moving mmap stuff to a feature gated `mmap` mod shouldn't be API changing, right? If not, then this should be ready to go and creating a `mmap` mod could be done in a point release.

Indeed, this should be fine, as long as we expose the public types in the same modules as we do now.

PR #134, opened 2020-04-28
Put mmap support in submodules to reduce number of feature gates

danieldk:
This change puts most mmap-related functionality in submodules, so
that we only have to feature gate the module, its use, and unit tests.

The storage wrappers are excluded, because the feature gating there is
applied to enum variants, which we can't really untangle.

### DIFF
--- a/src/chunks/storage/array.rs
+++ b/src/chunks/storage/array.rs
@@ -1,38 +1,44 @@
-#[cfg(feature = "memmap")]
-use std::fs::File;
-#[cfg(feature = "memmap")]
-use std::io::BufReader;
 use std::io::{Read, Seek, SeekFrom, Write};
 use std::mem::size_of;
 
-#[cfg(target_endian = "big")]
-use byteorder::ByteOrder;
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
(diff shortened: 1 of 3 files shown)


src/chunks/storage/array.rs
```diff
@@ -1,145 +1,147 @@
...
+
+    use super::NdArray;
+    use crate::chunks::io::{ChunkIdentifier, TypeId, WriteChunk};
+    use crate::chunks::storage::{Storage, StorageView};
+    use crate::error::{Error, Result};
+    use crate::util::padding;
+
+    /// Memory-mapped matrix.
+    #[derive(Debug)]
+    #[cfg(feature = "memmap")]
```
sebpuetz, line 34: Isn't this redundant since it's contained in the already feature-gated `mmap` mod?

danieldk: Indeed, missed that one!

danieldk:
> There's also a bunch of unused import warnings coming out of the cross compilation, any clue what's going on there?

Yes, some of the imports should be feature gated. I think these warning must have been present before this PR as well. But it seems like GitHub added a feature where they show them inline (which is nice!).

We could fix them in a separate PR. They only occur on big endian anyway.

PR #138, opened 2020-05-17
SubwordVocab: check that lengths can be represented by usize

danieldk:
finalfusion was developed with the assumption that it would run on
64-bit platforms and thus that embedding index computations would
never overflow. However, this can happen on x86, especially in subword
vocabs. Make this less likely to happen by enforcing two envariants:

- The upper bound of an indexer should by representable by usize.
- The size of the word vocabulary plus the upper bound of the indexer
  should be representable as usize.

### DIFF
--- a/src/chunks/vocab/subword.rs
+++ b/src/chunks/vocab/subword.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::convert::TryFrom;
 use std::io;
 use std::io::{Read, Seek, Write};
 use std::mem::size_of;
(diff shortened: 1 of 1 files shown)


danieldk:
Still not sure whether we should do this or only make this crate compile on 64-bit and 128-bit pointer sized platforms.

danieldk:
> Do you see clear downsides from supporting 32-bit this way? If no, then I'd say it's fine

That we currently don't test it, so there may be other things that need fixing.

I guess we don't have to do it for `u32` -> `usize` conversions. I hope nobody is trying to load embeddings on 16-bit platforms ;).

PR #143, opened 2020-06-04
Storage: allow lookup of multiple embeddings at the same time

danieldk:
There is some noise around the tests for the quantized embeddings. The amount of feature gating was getting a bit out of hand, so I moved all the tests that use memory-mapped quantized storage to a submodule.

### DIFF
--- a/src/chunks/storage/mod.rs
+++ b/src/chunks/storage/mod.rs
@@ -1,6 +1,6 @@
 //! Embedding matrix representations.
 
-use ndarray::{ArrayView2, ArrayViewMut2, CowArray, Ix1};
+use ndarray::{Array2, ArrayView2, ArrayViewMut2, CowArray, Ix1};
 
 mod array;
 #[cfg(feature = "memmap")]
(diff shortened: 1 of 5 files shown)


src/chunks/storage/mod.rs
```diff
@@ -23,6 +23,9 @@ pub use self::wrappers::{StorageViewWrap, StorageWrap};
 pub trait Storage {
     fn embedding(&self, idx: usize) -> CowArray<f32, Ix1>;
 
+    // Retrieve multiple embeddings.
+    fn embeddings(&self, indices: &[usize]) -> Array2<f32>;
```
sebpuetz, line 27: Did we intentionally not document the `Storage` trait? Should this be a doc-comment or just a inline comment?

danieldk: I think this is an accident :(. Changed to doc comment. We should do a separate PR to document these.

danieldk:
> Looks good, did you happen to run any benchs?

Not yet, but I'll do that once the next PR is done (which hooks this up to the lookup in `Embeddings`).

src/chunks/storage/mod.rs
danieldk: Fixed that. Good to go?

sebpuetz: Yup, good to go

PR #144, opened 2020-06-04
Embeddings::embedding: use Storage::embeddings

danieldk:
This makes lookups of unknown words faster, especially for quantized
embeddings.

Benchmarking results with differences larger than 5%:

Screenshot from 2020-06-04 12-04-55

* Lookups for `NdArray` became ~7-14% slower. Probably because of the extra allocation of the matrix by `Embeddings::storage`.
* Lookups for `MmapArray` became ~15-16% faster. Maybe the combined mmap -> `ArrayView` + select allows better optimization?
* As expected, lookups for `QuantizedArray` are much faster, ~164% for unknowns and ~60% allround.

I think it's worth it. Array lookups are extremely fast already, so the loss is barely noticable, whereas this will give a nice speedup for opq embeddings.

I guess there is another micro-optimization where we could use a regular lookup when the array of indices only has one element. But I guess that's too rare to be interesting.

### DIFF
--- a/src/embeddings.rs
+++ b/src/embeddings.rs
@@ -5,7 +5,7 @@ use std::iter::Enumerate;
 use std::mem;
 use std::slice;
 
-use ndarray::{Array1, ArrayViewMut1, CowArray, Ix1};
+use ndarray::{Array1, ArrayViewMut1, Axis, CowArray, Ix1};
 use rand::{RngCore, SeedableRng};
 use rand_xorshift::XorShiftRng;
 use reductive::pq::TrainPQ;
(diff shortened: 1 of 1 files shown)


sebpuetz:
Oh nice, I didn't think about using that for the subword lookups, I expected you'd add an `embeddings()` method on `Embeddings`: 
```Rust
fn embeddings(&self, batch: &[str]) -> Array2  {
...
}
```

But I guess there's some problems here with failed lookups: Return `None` if any lookup in the batch failed? Return `Err` in that case? Add a `default` param?

danieldk:
> But I guess there's some problems here with failed lookups: Return `None` if any lookup in the batch failed? Return `Err` in that case? Add a `default` param?

Such a method might indeed be worthwhile as well, but I agree that the semantics can be a bit difficult to nail down. Intuitively, I'd think something like `Vec<Option<Array1>>` would be nice to communicate failure, but then it's not as nice a representation as `Option<Array2>`.

PR #145, opened 2020-06-06
Embedding::{embedding_into, embedding_with_norm}: use Storage::embeddings

danieldk:


danieldk:
> It's a bit surprising to me that this is more efficient, if I understand correctly, `Array::select` copies the emeddings to a matrix, then we iterate over the rows of that array and sum into the target array. Why is this faster than directly summing from the embeddings into the target array?

It should only be more efficient for opq embeddings, since projection.

sebpuetz:
I see, guess the cost is negligible for arrays anyways.

PR #147, opened 2020-06-10
Add `euclidean_{distance,similarity}` to `WordSimilarityResult`

danieldk:
Does what is says on the tin (as long as we have unit vectors).

### DIFF
--- a/src/similarity.rs
+++ b/src/similarity.rs
@@ -32,6 +32,22 @@ impl<'a> WordSimilarityResult<'a> {
     pub fn cosine_similarity(&self) -> f32 {
         *self.similarity
     }
+
+    /// Get the euclidean distance between the vectors.
+    pub fn euclidean_distance(&self) -> f32 {
+        // Trivially derived from the law of cosines
+        (2f32 - 2f32 * self.cosine_similarity()).sqrt()
+    }
+
+    /// Returns the euclidean similarity.
+    ///
+    /// This method returns a similarity in *[0,1]*, where *0*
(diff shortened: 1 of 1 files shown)


danieldk:
> `analogy_by` and `similarity_by` take arbitrary `FnMut(ArrayView2<f32>, ArrayView1<f32>) -> Array1<f32>`, `embedding_similarity` can also produce non-cosine-similarity outputs.
> 
> This is not restricted to the `euclidean_distance` but also applies to `cosine_similarity` and `angular_similarity`, `WordSimilarityResult` only upholds that the similarity is `NotNan`, the methods could be misleading...but I don't think there's an easy fix and we've had it this way since the similarity module was added.

That's a great point! I think the proper solution would be to let the `by` function specify the return type.

Since probably a lesser-used part of the API, we can probably get away with making minor changes. I'll try to make proposal in a branch and then we can see if we want to do this.

danieldk:
Another possibility would be to make the `_by` functions private or remove them altogether.

As long as you are using unit vectors, all the similarity functions that you may want are trivially convertible. If you someone wants something really wacky, they always have access to the raw matrices.

So, I'd say: add a deprecation marker for the next minor release, remove completely thereafter.

sebpuetz:
> Another possibility would be to make the `_by` functions private or remove them altogether.
> 
> As long as you are using unit vectors, all the similarity functions that you may want are trivially convertible. If you someone wants something really wacky, they always have access to the raw matrices.
> 
> So, I'd say: add a deprecation marker for the next minor release, remove completely thereafter.

Sounds like the most sensible solution to me

PR #148, opened 2020-06-10
Deprecate the *By implementations for analogy/similarity

danieldk:
Couldn't deprecate the traits, since it gives compiler warnings and makes CI fail.

After this is merged, I'll add it to the `release-0.12` branch and then remove these traits from `master` completely.


danieldk:
> I guess l2-normalizing the input embedding in `embedding_similarity` would close the last hole, the cost should be negligible compared to the huge amount of dot-products in the actual similarity calculation.

That would be great, will you do a PR?

sebpuetz:
Sure, can do

PR #149, opened 2020-06-10
similarity: remove *By traits and implementations

danieldk:
It is unlikely that someone would use a custom similarity function
that cannot be trivially obtained from coside similarity. Removing
these traits simplifies the module and makes it easier to guarantee
certain invariants.

---

I guess another remaining issue is that the `WordSimilartyResult` fields are `pub` and someone can stuff it with their own values and the similarity methods may return nonsense. I guess we could solve this by adding assertions to these methods (ugly, but doesn't change the API) or break the API once more and make these fields private (and add a `word` method).

This again shows that one should always hide fields by default. Even for seemingly trivial data structures like this one, it bites you one day :cry: .

### DIFF
--- a/src/similarity.rs
+++ b/src/similarity.rs
@@ -4,7 +4,7 @@ use std::cmp::Ordering;
 use std::collections::{BinaryHeap, HashSet};
 use std::f32;
 
-use ndarray::{s, Array1, ArrayView1, ArrayView2, CowArray, Ix1};
+use ndarray::{s, ArrayView1, CowArray, Ix1};
 use ordered_float::NotNan;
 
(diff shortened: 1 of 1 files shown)


sebpuetz:
I don't think we use the `WordSimilarityResult` as input anywhere. So I don't think the fields being public is an issue at all for the API?

danieldk:
> I don't think we use the `WordSimilarityResult` as input anywhere. So I don't think the fields being public is an issue at all for the API?

Unless someone puts weird values in the similarity fields and then calls `angular_similarity`, `euclidean_distance`, etc. and hopes to get out something useful.

Not sure it that is a problem though ;).

PR #152, opened 2020-06-30
Add WriteFastText trait and implementation

danieldk:
First stab.

### DIFF
--- a/src/compat/fasttext/io.rs
+++ b/src/compat/fasttext/io.rs
@@ -1,7 +1,8 @@
 use std::convert::TryInto;
-use std::io::BufRead;
+use std::io::{BufRead, Write};
+use std::ops::Mul;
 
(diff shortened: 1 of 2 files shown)


danieldk:
> fastText reads/writes some different datatypes for e.g. the vocab item counts or the cfg values, these are represented as signed 32/64 bit integers. But I don't think this makes a difference for our purposes.

I think unsigned are the better representation, because for many of these values, negatives are actually invalid. But we should probably validate that the value is not larger than i32::MAX or i64::MAX after reading (if they were, the values were negative).

I will add another PR for that.

danieldk:
Just realized that that would perhaps clash with `finalfusion-python`, which uses -1 for unknown parameters? Hmpf. This is why people shouldn't put non-essential hyperparameters into the format (rather than into a metadata chunk like we do).

PR #153, opened 2020-07-01
WriteFastText: Set all non-essential hyperparams to 0 for simplicity

danieldk:


danieldk:
I'll tag a new minor release with these changes, so that we can integrate this in finalfrontier/ff-utils. We can add the additional `i32`/`i64` checks later.

PR #157, opened 2020-07-26
Update to reductive 0.5

danieldk:


danieldk:
Shuffling things around a bit to make BLAS use more flexible. We keep `netlib-src` underspecified. Then we can use `netlib-src` downstream with the default features reset and `system` enabled. Then we can link against any CBLAS/LAPACKE as long as it walks and quacks like one.

PR #158, opened 2020-07-27
Bucket to explicit vocab conversion.

sebpuetz:
Implement conversion from bucket subword vocabularies to explicitly
indexed subword vocabularies.

### DIFF
--- a/src/chunks/vocab/subword.rs
+++ b/src/chunks/vocab/subword.rs
@@ -351,6 +351,33 @@ where
 
         Ok(())
     }
+
+    /// Convert the hash-based vocabulary to an Explicit Vocabulary.
+    ///
+    /// N-grams in the range `(self.min_n..self.max_n)` are extracted from the words in the
+    /// vocabulary, each of these gets assigned an index from the `BucketIndexer` which is used to
+    /// determine the index in the explicit subword vocab.
+    pub fn to_explicit(&self) -> ExplicitSubwordVocab {
+        let mut ngram_index = HashMap::new();
+        let SubwordVocab {
+            words,
+            indices: _,
+            indexer,
+            min_n,
+            max_n,
+        } = &self;
+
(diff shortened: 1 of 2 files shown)


src/chunks/vocab/subword.rs
```diff
@@ -351,6 +351,36 @@ where
...
+        let mut ngram_set = HashSet::new();
+        let mut ngrams = Vec::new();
+        let SubwordVocab {
+            words,
+            indices: _,
+            indexer,
+            min_n,
+            max_n,
+        } = &self;
+        let bracketed_words = words.iter().map(Self::bracket).collect::<Vec<_>>();
```
danieldk, line 370: Does this need to be collected rather than directly iterating over this?

sebpuetz: We have the choice to collect the bracketed words or to allocate the ngrams twice as `String`. I'd expect fewer words than ngrams in a vocab, so this might be cheaper...

danieldk: But you can check whether an n-gram has been seen by making `ngrams` a `HashMap`. `new_with_indices` is not flexible enough to take a `HashMap`, but you can probably change that constructor to take `IntoIterator` without breaking the API?

sebpuetz: Not sure whether
```Rust
if !ngram_index.contains_key(ngram) {
    ngram_index.insert(ngram.to_string(), idx);
}
```
or the `Entry` API makes more sense here.

danieldk: The `Entry` API requires only one lookup, but possibly requires more allocations (if this isn't optimized away).

Since this is just a one-time conversion, I'd probably go for the `Entry` API, since it is a bit more readable.

sebpuetz: That was also my thought, although I didn't consider the extra allocation might disappear!

PR #160, opened 2020-07-29
Implement conversion to explicit subword embeddings.

sebpuetz:
There is a small public API change included:

`ExplicitIndexer::new_with_indices()` now also returns the `old_to_new`mapping. Otherwise it's much more involved to figure out which bucket was moved to which index.

I'll add a new `pub (crate)` constructor if you prefer to keep the API stable.

### DIFF
--- a/src/chunks/storage/mod.rs
+++ b/src/chunks/storage/mod.rs
@@ -29,6 +29,20 @@ pub trait Storage {
     fn shape(&self) -> (usize, usize);
 }
 
+pub(crate) mod sealed {
+    use crate::chunks::storage::Storage;
+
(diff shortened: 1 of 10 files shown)


src/chunks/storage/mod.rs
```diff
@@ -29,6 +29,20 @@ pub trait Storage {
     fn shape(&self) -> (usize, usize);
 }
 
+pub(crate) mod sealed {
+    use crate::chunks::storage::Storage;
+
+    /// Return a new storage Storage based on the mapping.
```
danieldk, line 35:
```suggestion
    /// Return a new storage from an existing Storage based on a mapping.
```

PR #165, opened 2020-08-20
Fix empty subword indices.

### LINKED ISSUE #164: Indexing a BucketVocab with the empty String returns Some
With range 3-6 these fail:

~~~Rust
fn empty_string_idx(voc: BucketSubwordVocab) {
    assert!(voc.idx("").is_none())
}

fn empty_string_embed(e: Embeddings<BucketSubwordVocab, NdArray>) {
    assert!(e.embedding("").is_none())
}
~~~

I'd expect the same for other setups, e.g. range 4-6 and length 1 inputs will probably behave the same.

sebpuetz:
Closes #164


danieldk (approved):
Good catch!

PR #176, opened 2021-07-12
Switch from memmap to memmap2

djc:
memmap is no longer maintained, and the RustSec advisory DB issues a warning
when it appears in your dependency graph. memmap2 is a fork that is being
maintained, and is substantially more popular than the other suggested
alternative.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,12 +28,13 @@ serde = { version = "1", features = ["derive"] }
 thiserror = "1"
 toml = "0.5"
 
-[dependencies.memmap]
-version = "0.7"
+[dependencies.memmap2]
+version = "0.3"
 optional = true
 
 [features]
 default = ["memmap"]
+memmap = ["memmap2"]
 
 [dev-dependencies]
 approx = "0.4"
@@ -53,4 +54,3 @@ harness = false
 [[bench]]
 name = "quantized"
 harness = false
-
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,6 +15,10 @@ jobs:
           override: true
       - uses: actions-rs/cargo@v1
         with:
+          command: check
+          args: --all-targets
+      - uses: actions-rs/cargo@v1
+        with:
           command: test
 
   test-cross:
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -28,12 +28,12 @@ serde = { version = "1", features = ["derive"] }
...
 
-[dependencies.memmap]
-version = "0.7"
+[dependencies.memmap2]
+version = "0.3"
 optional = true
 
 [features]
-default = ["memmap"]
+default = ["memmap2"]
```
danieldk, line 36: This would break the 'memmap' feature gating that is used in some module. I propose that we add an indirection along the lines of:

```
[features]
default = ["memmap"]
memmap = ["memmap2"]
```

src/chunks/storage/wrappers.rs
```diff
@@ -186,8 +186,6 @@ impl WriteChunk for StorageWrap {
         match self {
             #[cfg(all(feature = "memmap", target_endian = "little"))]
             StorageWrap::MmapArray(inner) => inner.chunk_identifier(),
-            #[cfg(target_endian = "big")]
-            StorageWrap::MmapArray(_inner) => unimplemented!(),
```
danieldk, line 190: Why is the big-endian gating removed?

danieldk (changes requested):
Thanks for this change! I have added one suggested change and a question.

Cargo.toml
djc: Done.

src/chunks/storage/wrappers.rs
djc: I got some warnings somewhere, probably due to the feature gating issues.

danieldk: We do not support memory mapping for big-endian architectures. It will work, but you will get incorrect embeddings, since the finalfusion format stores data in little-endian format.

danieldk:
I rebased on master, that should fix the Clippy errors.

danieldk:
Dropping 403ecc9213b21f1fb7c070c10f04a6adba73b7ab would resolve the Clippy warnings.

djc:
Dropped the clippy fixes.

danieldk (approved):
LGTM, thanks!

PR #180, opened 2021-11-07
Add support for Floret embeddings

danieldk:
Add support for reading and using Floret embeddings. Floret combines subword
embeddings with Bloom embeddings:

https://github.com/explosion/floret#how-floret-works

This allows Floret embedding matrices to be compact like Bloom embeddings,
while also using subword units.

Supporting Floret embeddings requires some architectural changes to
finalfusion:

1. ngrams -> index mappings are many-to-many

   finalfusion assumed that each n-gram maps to a single index. In Floret
   embeddings, an n-gram can map to multiple embeddings. finalfusion is
   extended in this change to support returning multiple indices for an
   n-gram.

2. Floret also hashes the word itself, not just the subwords. This change
   adds a boolean flag. If it is enabled, the word itself is also hashed,
   and its index is returned.

3. Floret permits different begin/end-of-word markers. These are made
   configurable for subword vocabs.

### DIFF
--- a/src/chunks/vocab/subword.rs
+++ b/src/chunks/vocab/subword.rs
@@ -5,19 +5,23 @@ use std::io::{Read, Seek, Write};
 use std::mem::size_of;
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
+use smallvec::SmallVec;
 
 use crate::chunks::io::{ChunkIdentifier, ReadChunk, WriteChunk};
 use crate::chunks::vocab::{create_indices, read_vocab_items, write_vocab_items, Vocab, WordIndex};
 use crate::compat::fasttext::FastTextIndexer;
+use crate::compat::floret::FloretIndexer;
 use crate::error::{Error, Result};
 use crate::subword::{
-    BucketIndexer, ExplicitIndexer, FinalfusionHashIndexer, Indexer,
+    BucketIndexer, ExplicitIndexer, FinalfusionHashIndexer, Indexer, StrWithCharLen,
     SubwordIndices as StrSubwordIndices,
 };
-use crate::util::CollectWithCapacity;
 
 /// fastText vocabulary with hashed n-grams.
 pub type FastTextSubwordVocab = SubwordVocab<FastTextIndexer>;
+
+/// floret vocabulary with hashed n-grams.
+pub type FloretSubwordVocab = SubwordVocab<FloretIndexer>;
 
 /// finalfusion vocabulary with hashed n-grams.
 pub type BucketSubwordVocab = SubwordVocab<FinalfusionHashIndexer>;
@@ -25,9 +29,22 @@ pub type BucketSubwordVocab = SubwordVocab<FinalfusionHashIndexer>;
 /// finalfusion vocabulary with explicit n-grams.
 pub type ExplicitSubwordVocab = SubwordVocab<ExplicitIndexer>;
 
+/// What to compute subwords for.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum IndicesScope {
+    /// Only subword indices.
+    Subwords,
+
+    /// Word and subword indices.
+    WordAndSubword,
+}
+
 /// Vocabulary with subword units.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct SubwordVocab<I> {
+    bow: String,
+    eow: String,
+    indices_scope: IndicesScope,
     indexer: I,
     indices: HashMap<String, usize>,
     words: Vec<String>,
@@ -39,17 +56,51 @@ impl<I> SubwordVocab<I>
 where
     I: Indexer,
 {
-    const BOW: char = '<';
-    const EOW: char = '>';
+    const DEFAULT_BOW: &'static str = "<";
+    const DEFAULT_EOW: &'static str = ">";
 
-    /// Construct a new `SubwordVocab`.
+    /// Construct a new `SubwordVocab` with default word boundaries.
     ///
     /// Words are assigned indices in the given order. NGrams in range `(min_n..max_n)` are
     /// considered. The `indexer` is used to look up indices for the NGrams produced by this
     /// `SubwordVocab`.
     ///
     /// Panics when there are duplicate words.
-    pub fn new(words: impl Into<Vec<String>>, min_n: u32, max_n: u32, indexer: I) -> Self {
+    pub fn new(
+        words: impl Into<Vec<String>>,
+        min_n: u32,
+        max_n: u32,
+        indexer: I,
+        indices_scope: IndicesScope,
+    ) -> Self {
+        Self::new_with_boundaries(
+            words,
+            min_n,
+            max_n,
+            indexer,
+            indices_scope,
+            Self::DEFAULT_BOW,
+            Self::DEFAULT_EOW,
+        )
+    }
+
+    /// Construct a new `SubwordVocab`.
+    ///
+    /// Words are assigned indices in the given order. NGrams in range `(min_n..max_n)` are
+    /// considered. The `indexer` is used to look up indices for the NGrams produced by this
+    /// `SubwordVocab`. N-grams are extracted from words with the `bow` and `eow` boundaries
+    /// added.
+    ///
+    /// Panics when there are duplicate words.
+    pub fn new_with_boundaries(
+        words: impl Into<Vec<String>>,
+        min_n: u32,
+        max_n: u32,
+        indexer: I,
+        indices_scope: IndicesScope,
+        bow: impl Into<String>,
+        eow: impl Into<String>,
+    ) -> Self {
         let words = words.into();
         let indices = create_indices(&words);
         assert_eq!(
@@ -74,6 +125,9 @@ where
         );
 
         SubwordVocab {
+            bow: bow.into(),
+            eow: eow.into(),
+            indices_scope,
             indexer,
             indices,
             words,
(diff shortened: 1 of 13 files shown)


danieldk:
@sebpuetz I have put this in draft, since I need to update the documentation in various places and clean up some code. But I would like to check if you are principally ok with the changes.

Floret embeddings use a 1-to-many mapping for n-grams. I have extended the subword indecing interface, so that we do not return `Option<u64>`, but a vector of `u64`. An empty vector then means that that no index could be found. To avoid making many additional heap allocations, I store the vector of `u64` as `SmallVec`, so that we still rely on stack allocation.

We could follow a completely different route here and not extend the `SubwordVocab` interface and keep the 1-to-{0, 1} mapping between n-grams and indices and create a separate `FloretVocab` vocabulary.

Let me know what your thoughts are.

src/compat/floret/indexer.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+}
+
+impl Indexer for FloretIndexer {
+    fn index_ngram(&self, ngram: &StrWithCharLen) -> NGramVec {
+        let hash = murmur3_x64_128(&mut Cursor::new(ngram.as_bytes()), self.seed)
+            .expect("Murmur hash failed");
+
+        let hash_array: [u32; 4] = unsafe { std::mem::transmute(hash) };
```
sebpuetz, line 37: Feel free to ignore this if you think it's too pedantic, but I think a small comment regarding safety is nice to have next to an unsafe block.

On a different note, the mips64 tests failed while other platforms worked, does this `transmute` lead to endianness issues?

src/chunks/vocab/subword.rs
```diff
@@ -173,25 +214,30 @@ where
 pub trait SubwordIndices {
     /// Return the subword indices of the subwords of a string,
     /// according to the subword vocabulary.
-    fn subword_indices(&self, word: &str) -> Option<Vec<usize>>;
+    fn subword_indices(&self, word: &str, hash_word: bool) -> Option<Vec<usize>>;
```
sebpuetz, line 217: While looking at the trait definition, the `bool` param is fine, but at calling sites it's not really self-explanatory what `true` or `false` as the second argument to this function means. Could be nicer and more explicit to use an enum? Although I don't have a strong opinion on that.

Still, a short explanation of the expected behaviour given this param in the docs would be great!

sebpuetz (commented):
I think it's fine to return an empty `SmallVec` to indicate absence of indices and I don't think we need to implement an entirely different Vocab kind. 

I believe there might be some endianness issue going on with the transmute in the `index_ngram` impl.

Besides that, the `hash_word: bool` param on `SubwordIndices::subword_incides` is not really self-explanatory in contexts where that method is called. It's mostly internal, but it's still an external API and at least the param should be explained!

adrianeboyd:
On big endianness, murmurhash is the issue. See this recent discussion: https://github.com/explosion/spaCy/issues/9428#issuecomment-948182219

danieldk:
> On big endianness, murmurhash might be the issue. See this recent discussion: [explosion/spaCy#9428 (comment)](https://github.com/explosion/spaCy/issues/9428#issuecomment-948182219)

Thanks for the pointer! That indeed seems to be the primary issue.

src/chunks/vocab/subword.rs
danieldk: Using an enum here is a great idea!

sebpuetz:
Are you sure? I tested this with cross and got it to work with changes in the indexer fn:

~~~ Rust
diff --git a/src/compat/floret/indexer.rs b/src/compat/floret/indexer.rs
index 4d9918d..c01d5fe 100644
--- a/src/compat/floret/indexer.rs
+++ b/src/compat/floret/indexer.rs
@@ -1,3 +1,4 @@
+use std::convert::TryInto;
 use std::io::Cursor;
 
 use murmur3::murmur3_x64_128;
@@ -33,16 +34,16 @@ impl Indexer for FloretIndexer {
     fn index_ngram(&self, ngram: &StrWithCharLen) -> NGramVec {
         let hash = murmur3_x64_128(&mut Cursor::new(ngram.as_bytes()), self.seed)
             .expect("Murmur hash failed");
-
-        let hash_array: [u32; 4] = unsafe { std::mem::transmute(hash) };
+        let le_bytes = hash
+            .to_le_bytes();
+        let le_bytes = le_bytes
+            .chunks_exact(4)
+            .map(|c| u32::from_le_bytes(c.try_into().unwrap()));
 
         let mut indices = smallvec![0; self.n_hashes as usize];
-        for i in 0..self.n_hashes as usize {
-            //hashes[i as usize] = ((hash & 0xffffffff) as u64) % self.n_buckets;
-            indices[i] = hash_array[i] as u64 % self.n_buckets;
-            //eprintln!("hash {}: {} , index: {}", i, (hash & 0xffffffff) as u64, hashes[i as usize]);
-            //hash >>= 32;
-        }
+        indices.iter_mut().zip(le_bytes).for_each(|(idx, hash)| {
+            *idx = hash as u64 % self.n_buckets;
+        });
 
         indices
     }
~~~

~~~sh
cross test --target mips64-unknown-linux-gnuabi64g
~~~

Fails without the patch

src/chunks/vocab/subword.rs
```diff
@@ -97,11 +133,11 @@ where
         self.max_n
     }
 
-    fn bracket(word: impl AsRef<str>) -> String {
+    fn bracket(&self, word: impl AsRef<str>) -> String {
         let mut bracketed = String::with_capacity(word.as_ref().len() + 2);
```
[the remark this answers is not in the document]
danieldk: Thanks, fixed!

src/chunks/vocab/subword.rs
danieldk: I added the following enum:

```
pub enum IndicesScope {
    Subwords,
    WordAndSubword,
```

Looks far more readable at various call sites.

src/compat/floret/indexer.rs
danieldk: We should probably get rid of the `unsafe` altogether. A simple bit shift approach get compiled away:

https://godbolt.org/z/KjKnKbvPM

But I have to investigate the ramifications a bit wrt. murmur3 and endianness.

src/chunks/vocab/subword.rs
sebpuetz: Yep, looks nice!

danieldk:
Right, it could be that it's as simple as byte swapping the output.

sebpuetz:
Iterators also compile to the same instructions:

https://godbolt.org/z/hEbYMKG1W

src/compat/floret/indexer.rs
sebpuetz: I compared some outputs of the Murmur3 implementation and they are the same for any inputs I tried on mips via cross and on native x86_64:

~~~Rust
use std::io::Cursor;

fn main() {
    let inputs = vec!["test", "another", "asd"];
    for input in inputs {
        println!(
            "{}: {}",
            input,
            murmur3::murmur3_x64_128(&mut Cursor::new(input.as_bytes()), 2166136261).unwrap()
        );
    }
    // mips
    // test: 143416175659979471159943107701931823025
    // another: 19454989077232334339732095908570983208
    // asd: 49449963734955912795853045277450408020

    // x86_64
    // test: 143416175659979471159943107701931823025
    // another: 19454989077232334339732095908570983208
    // asd: 49449963734955912795853045277450408020

    let input = (0..255).cycle().take(1024).collect::<Vec<u8>>();
    println!(
        "{}",
        murmur3::murmur3_x64_128(&mut Cursor::new(input), 2166136261).unwrap()
    );
    // x86_64: 121525936877850996829446231388818626050
    // mips: 121525936877850996829446231388818626050
}
~~~

danieldk:
I investigated a bit more. The Rust version of `murmur3_x64_128` is actually correct wrt. endianness. The same integer is returned on LE and BE, but just with their respective endianness encodings. Floret in spaCy does the Cython equivalent of a transmute:

https://github.com/explosion/spaCy/blob/86af0234abfd2d3d2a609bd46eaa4b475477eb3b/spacy/vectors.pyx#L359

So, the 128-but number is transmuted into four 32-bit unsigned. Suppose that the hash is *158233910618652003523339446073433302185*, then:

```
LE: [169, 164, 144, 244, 167, 125, 125, 100, 53, 152, 5, 122, 31, 190, 10, 119]
BE: [119, 10, 190, 31, 122, 5, 152, 53, 100, 125, 125, 167, 244, 144, 164, 169]
```

So, transmuting on BE is then incorrect (as you suspected earlier @sebpuetz), because on the first `u32` is the least significant `u32` on LE and the most significant `u32` on BE. One straightforward way of working around this is bit shifting + casting, since it will always give you the least significant bytes first.

On LE, this will be compiled to a simple transmute. On BE (depending on the instruction set), it basically swaps the 64-bit words and then within each word rotates the bits by 32 (which reverses the order of `u32`s).

sebpuetz:
If you're after a micro optimization, I think the iterators have a bounds check less than the explicit for-loop https://godbolt.org/z/EvPx8h8G8

Other than that, I think this PR looks good - apart from the pedantic field order lints ;)

danieldk:
> If you're after a micro optimization, I think the iterators have a bounds check less than the explicit for-loop https://godbolt.org/z/EvPx8h8G8 - probably not measurable and doesn't actually matter

Yeah, and the big question these days is: what does it look like on AArch64 :). Any way, I think here the range for loop is clearer, so we should probably use that.

> Otherwise I think this PR looks good - apart from the pedantic field order lints ;)

Cool. I think it's almost ready for going out of draft.

danieldk:
Geez clippy, I don't think I like your suggestion, but this time you win. 📎

src/chunks/vocab/subword.rs
```diff
@@ -497,15 +587,29 @@ where
...
                 io::ErrorKind::Other.into(),
-            )
-        })?;
+            ));
+        }
+
+        if indices.len() > 1 {
+            return Err(Error::io_error(
+                format!(
+                    "Indexer maps n-gram to multipli indices during serialization: {}",
```
sebpuetz, line 604:
```suggestion
                    "Indexer maps n-gram to multiple indices during serialization: {}",
```

Typo!

sebpuetz (approved):
Small typo in an error message but otherwise looked good to me

Are you planning on adding a finalfusion chunk for the vocab, too?

danieldk:
Cool, thanks for reviewing! I'll merge this when CI is green.

I guess we should follow this up elsewhere, but we may want to add a new vocab chunk for Bloom-family of embeddings (spaCy-like, floret), so that we can also store them natively.

Would also be cool if we can train floret embeddings with finalfrontier, but I didn't really look at the ramifications yet.

PR #181, opened 2021-11-08
Add support for writing floret embeddings

danieldk:


src/compat/floret/io.rs
```diff
@@ -106,10 +108,52 @@ impl ReadFloretText for Embeddings<FloretSubwordVocab, NdArray> {
...
+/// Write embeddings in the floret format.
+pub trait WriteFloretText
+where
+    Self: Sized,
+{
+    /// Read embeddings in the floret format.
+    fn write_floret_text(&self, write: &mut dyn Write) -> Result<()>;
+}
+
+impl WriteFloretText for Embeddings<FloretSubwordVocab, NdArray> {
```
sebpuetz, line 120: I haven't looked into this, so this is just a genuine question: Are there compatibility issues between our quantized storage interface and floret embeddings?

sebpuetz (approved):
Appears to be doing what you're expecting it to do

src/compat/floret/io.rs
danieldk: I don't think so. Since the quantized storage is oblivious to the vocabulary, it should just work if we have a Vocab chunk for floret in place.

Would be cool, a small embedding matrix thanks to bloom embeddings, compressed even further through quantization.

PR #182, opened 2021-11-08
Support FloretSubwordVocab in VocabWrap

danieldk:
Support for writing FloretSubwordVocab chunks is currently not
implemented.

### DIFF
--- a/src/embeddings.rs
+++ b/src/embeddings.rs
@@ -273,8 +273,8 @@ where
 {
     /// Try to convert to explicitly indexed subword embeddings.
     ///
-    /// Conversion fails if the wrapped vocabulary is `SimpleVocab` or already an
-    /// `ExplicitSubwordVocab`.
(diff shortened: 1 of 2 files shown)


src/embeddings.rs
```diff
@@ -289,6 +289,10 @@ where
             VocabWrap::ExplicitSubwordVocab(_) => {
                 Err(Error::conversion_error("ExplicitVocab", "ExplicitVocab"))
             }
+            VocabWrap::FloretSubwordVocab(_) => Err(Error::conversion_error(
+                "FloretSubwordVocab",
+                "ExplicitVocab",
+            )),
```
sebpuetz, line 295: Not part of the changed lines so I'm commenting down here: The documentation is out of sync with this change since the floret vocab doesn't support (in the current state) conversion to an explicit subword vocab

sebpuetz (dismissed):
Just a small doc fix, otherwise should be fine

src/embeddings.rs
danieldk: Fixed.

PR #187, opened 2021-11-12
Improve indexing scope handling

danieldk:
This PR combines two changes:

* Move down `IndicesScope` handling to `SubwordIndices` trait. This is a better place to do this, because e.g. finalfrontier uses the `SubwordIndices` trait.
* Add the `Indexer::scope` method. This allows us to get the indexing scope from the indexer, rather than passing it around everywhere.

### DIFF
--- a/src/subword.rs
+++ b/src/subword.rs
@@ -375,6 +375,16 @@ impl<'a> Iterator for NGrams<'a> {
     }
 }
 
+/// What to compute subwords for.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum IndicesScope {
+    /// Only substring indices.
+    Substrings,
+
+    /// String and substring indices.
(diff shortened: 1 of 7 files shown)


src/subword.rs
```diff
@@ -463,6 +497,11 @@ where
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
+        if let Some(string) = self.string {
+            self.string = None;
```
sebpuetz, line 501:
```suggestion
        if let Some(string) = self.string.take() {
```
Nitpicking here, so feel free to ignore

sebpuetz (approved):

src/subword.rs
danieldk: 👍 fixed

sebpuetz (approved):

PR #189, opened 2021-11-25
Add a finalfusion chunk for floret vocabs

danieldk:


src/error.rs
```diff
@@ -43,6 +47,13 @@ pub enum Error {
 }
 
 impl Error {
+    pub fn context(context: impl Into<String>, error: impl Into<Box<Error>>) -> Self {
+        Error::Context {
+            context: context.into(),
+            error: error.into(),
+        }
+    }
+
```
sebpuetz, line 56: Not that important, but I think an associated builder-like function could be pretty ergonomic:
```suggestion
    pub fn context(self, context: impl Into<String>) -> Self {
        let error = Box::new(self);
        Error::Context {
            context: context.into(),
            error: error,
        }
    }

```

sebpuetz (approved):
Looks good!

src/error.rs
danieldk: That's much better, thanks!

PR #190, opened 2021-11-30
Update to reductive 0.8

danieldk:


danieldk:
Merging this without review, only renames some types to match changes in `reductive`.

PR #191, opened 2021-11-30
Add batch embedding lookups

danieldk:
This change adds the `embedding_batch` and `embedding_batch_into` methods.
These methods look up the embeddings for multiple words. The motivation
is threefold:

1. It provides the convenience of looking up multiple embeddings at
   once.
2. Words which occur more than once in the batch only need to be looked
   up once.
3. These method can avoid some allocations if the downstream user is
   concatenating the embeddings in a batch anyway.

---

For multiple occurrences of a word, the embedding is first realized in-place for the first occurrence of that word and then copied into the other rows in which the word occurs. I am not sure if using `multi_slice_mut` here is too convoluted. It avoids an allocation in the case of a ~known word~, since it will be copied into the output matrix in-place. ~Currently, I don't think it provides a benefit for unknown words, since for those we first look up multiple embeddings and sum then anyway.~

@sebpuetz let me know what you think. If we think it's premature implementation, I'll revert the implementation to what I had earlier (fetch the embedding into a temporary and then copy that temporary into all rows).

### DIFF
--- a/src/embeddings.rs
+++ b/src/embeddings.rs
@@ -1,11 +1,12 @@
 //! Word embeddings.
 
+use std::collections::HashMap;
 use std::io::{Read, Seek, Write};
 use std::iter::Enumerate;
 use std::mem;
 use std::slice;
 
-use ndarray::{Array1, ArrayViewMut1, Axis, CowArray, Ix1};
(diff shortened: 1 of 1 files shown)


danieldk:
@sebpuetz any objections against merging this?

sebpuetz (approved):
No objections

PR #192, opened 2021-12-03
Update to reductive 0.9

danieldk:


danieldk:
Merging this without review, since it's only making compat changes for reductive 0.9.

PR #193, opened 2021-12-03
Improve error handling

danieldk:
- `Error::context`: mark the inner error as the error source. This
  allows a downstream error handler (e.g. anyhow) to format the error
  chain nicely.
- Rename `Error::Shape` to `Error::MatrixShape` to narrow down the scope
  of the variant. Mark the shape error as the error source.
- Split `Error::Io` into `Error::{read, write}`. Mark the wrapped
  `io::Error` as the error source.
- Do not use debug representations of values in errors, we want their
  string representations instead.
- Remove the `Error` suffix from some variants. They are part of an
  error enum, we know that they are errors :^).

### DIFF
--- a/src/chunks/io.rs
+++ b/src/chunks/io.rs
@@ -34,7 +34,7 @@ impl ChunkIdentifier {
     {
         let chunk_id = read
             .read_u32::<LittleEndian>()
-            .map_err(|e| Error::io_error("Cannot read chunk identifier", e))?;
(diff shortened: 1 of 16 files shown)


danieldk:
It probably makes sense to cut a release after this PR to get the new floret support out. I'll check if all the dependencies are still up to date.

sebpuetz (approved):

PR #197, opened 2021-12-17
Add a batch size option to similarity and anology queries

danieldk:
This allows the user to trade-off efficiency for lower memory use.

@sebpuetz I am not sure if we want to break the current API or add new methods for this functionality (like `similarity_batched`).

In favor of adding new trait methods:

* The API does not change.

Against adding new trait methods:

* `Analogy` and `EmbeddingSimilarity` already have 2 methods, this would double the number to 4.

### DIFF
--- a/src/compat/fasttext/io.rs
+++ b/src/compat/fasttext/io.rs
@@ -620,7 +620,7 @@ mod tests {
     #[test]
     fn test_read_fasttext() {
         let embeddings = read_fasttext();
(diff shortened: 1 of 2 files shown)


sebpuetz (approved):
I haven't used these APIs since that workshop presentation in 2019, I don't have an opinion on these changes.

If you need the changes for your Python bindings, then this change is fine with me.

The implementation looks correct and although the additional parameter breaks API compatibility it's probably okay to do so.

PR #199, opened 2023-10-09
Update dependencies

djc:
Would be nice to get this in a release.

Given the lack of activity, what will maintenance be like going forward?


sebpuetz:
> Given the lack of activity, what will maintenance be like going forward?

I'm happy to review changes / dependency updates, I haven't been using the library (or rather anything in the project) for a couple of years so nothing has been driving active development.

I'll try to go through the other repos during the week and take care of some updates.

djc:
Thanks for the quick response!

djc:
Looks like the mips64 target is not available for Rust 1.72: https://github.com/rust-lang/rust/issues/115218.

Let me know how you want this handled?

sebpuetz:
> Looks like the mips64 target is not available for Rust 1.72: [rust-lang/rust#115218](https://github.com/rust-lang/rust/issues/115218).
> 
> Let me know how you want this handled?

After a cursory look at the issue, it looks like RustCrypto is using PPC32 as an alternate BE target: https://github.com/RustCrypto/sponges/pull/62

I believe our intention was to test BE targets, too, so that should be a viable alternative

djc:
I added a version bump to 0.18, since some of the dependencies that got bumped are visible in the public API.

danieldk:
> I'm happy to review changes / dependency updates, I haven't been using the library (or rather anything in the project) for a couple of years so nothing has been driving active development.

Same here, I don't use finalfusion/finalfrontier anymore, but am happy to review stuf

src/chunks/metadata.rs
```diff
@@ -16,18 +15,18 @@ use crate::io::ReadMetadata;
 /// finalfusion metadata in TOML format.
 #[derive(Clone, Debug, PartialEq)]
 pub struct Metadata {
-    inner: Value,
+    inner: toml::Table,
```
danieldk, line 18: Any reason not to import `toml::Table` above?

src/compat/fasttext/io.rs
```diff
@@ -107,7 +105,7 @@ impl ReadFastTextPrivate for Embeddings<FastTextSubwordVocab, NdArray> {
             )));
         }
 
-        let metadata = Value::try_from(config).map_err(|e| {
+        let metadata = toml::Table::try_from(config).map_err(|e| {
```
danieldk, line 108: Same.

src/chunks/metadata.rs
djc: Changed it to import the type.

sebpuetz (approved):
LGTM

danieldk (approved):

djc:
Thanks! Are there any other blockers to getting this published on crates.io?

danieldk:
> Thanks! Are there any other blockers to getting this published on crates.io?

Nope, I'll get to it now.

danieldk:
Done, thanks for making this happen!

Issue #198, opened 2023-06-05
Reading GoogleNews word2vec model fails


victorlaurentiu:
Trying to read the GoogleNews-vectors-negative300.bin word2vec model triggers this assert:
https://github.com/finalfusion/finalfusion-rust/blob/main/src/chunks/vocab/simple.rs#L28
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `3000000`,
  right: `2999997`: words contained duplicate entries.'
```
(when constructing a new simple vocabulary, the number of indices (3,000,000) ends up different than the number of words (2,999,997))

After some investigations I removed this word trimming and it worked fine afterwards:
https://github.com/finalfusion/finalfusion-rust/blob/main/src/compat/word2vec.rs#L98

I assume the model contains tokens that get trimmed into the same words.

The model is from: https://code.google.com/archive/p/word2vec
Code:
```rust
let mut reader = BufReader::new(File::open("GoogleNews-vectors-negative300.bin").unwrap());
let model = Embeddings::read_word2vec_binary(&mut reader).unwrap();
```

danieldk:
Thank you for reporting this and sorry for the late reply. I think we added the trimming as a precaution and it should probably be removed.

PR #202, opened 2025-02-22
Do not trim words as it breaks the popular googlenews model

TomMD:
See #198


TomMD:
Reupping this pr after two weeks. I build from my branch but am sure in two years I'll forget and try to use the upstream cargo package again.

TomMD:
Ping. We are at the year mark and I'm guessing there are no e-mail alerts for activity here.

